PR #14, opened 2018-10-12
Court report expand courtid and court report endpoints

andymarke:


src/main/java/uk/gov/justice/digital/delius/controller/CourtReportController.java
```diff
@@ -0,0 +1,113 @@
...
+    private ResponseEntity<List<CourtReport>> courtReportsResponseEntityOf(Optional<Long> maybeOffenderId) {
+        return maybeOffenderId
+            .map(offenderId -> new ResponseEntity<>(courtReportService.courtReportsFor(offenderId), OK))
+            .orElseGet(this::courtReportsNotFound);
+    }
+
+    private ResponseEntity<CourtReport> courtReportResponseEntityOf(Optional<Long> maybeOffenderId,  Long courtReportId) {
+        Optional<CourtReport> maybeCourtReport = maybeOffenderId.flatMap(offenderId -> courtReportService.courtReportFor(offenderId, courtReportId));
+        return maybeCourtReport.map(
+                courtReport -> new ResponseEntity<>(courtReport, OK)).orElse(courtReportNotFound());
```
retroredge: Any reason to use orElse here bit orElseGet above? Either is OK in this case but I've been bitten by NPE when using orElse in the past with more complex payloads

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/CourtReport.java
```diff
@@ -114,8 +115,9 @@
     @Column(name = "OFFENDER_ID")
     private Long offenderId;
 
-    @Column(name = "REQUIRED_BY_COURT_ID")
-    private Long requiredByCourtId;
+    @JoinColumn(name = "REQUIRED_BY_COURT_ID", referencedColumnName = "COURT_ID")
```
retroredge: Mike, oddly it worked without 'referencedColumnName' but it feels like it documents the model better with it in

src/test/java/uk/gov/justice/digital/delius/CourtReportAPITest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    private String aValidToken() {
+        return aValidTokenFor(UUID.randomUUID().toString());
+    }
+
+    private String aValidTokenFor(String distinguishedName) {
+        return "Bearer " + jwt.buildToken(UserData.builder()
+                .distinguishedName(distinguishedName)
+                .uid("bobby.davro").build());
+    }
```
retroredge: Could also test the failed auth scenarios. IIRC we do this in the other controller tests

andymarke: Yep, added scenarios for missing token

src/main/java/uk/gov/justice/digital/delius/controller/CourtReportController.java
andymarke: the notFound() methods don't do anything that would be any more expensive than constructing a new lambda function so probably just not worth it. I would usually use orElseGet if it was something that would be destructive if we executed that code or very expensive.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/CourtReport.java
llmikeyj: Yes - because we are joining on a one to one the assumption is that we're joining on the primary key. This is also very confusing /  annoying when trying to join on otherwise unique (but not primary id) columns

src/test/java/uk/gov/justice/digital/delius/CourtReportAPITest.java
```diff
@@ -0,0 +1,326 @@
...
+    public void missingOffenderRecordResultsInAllReportsForOffenderByCrnNotFound() {
+        when(offenderService.offenderIdOfCrn("CRN1")).thenReturn(Optional.empty());
+
+        given()
+            .header("Authorization", aValidToken())
+            .when()
+            .get("offenders/crn/CRN1/courtReports")
+            .then()
+            .statusCode(404);
+
```
llmikeyj: Nick, did you put a check in here to ensure that we don't get a false positive for 404

PR #15, opened 2018-10-24
Personal Circumstance for offender

andymarke:


src/main/java/uk/gov/justice/digital/delius/data/api/PersonalCircumstance.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Data;
+
+import java.time.LocalDateTime;
+
+@Data
+@Builder
+public class PersonalCircumstance {
+    private Long personalCircumstanceId;
+    private Long offenderId;
+    private KeyValue type;
```
andymarke: Should these be called `type` and `subType` or `personalCircumstanceType` etc ?

src/main/java/uk/gov/justice/digital/delius/service/PersonalCircumstanceService.java
```diff
@@ -0,0 +1,37 @@
...
+        this.personalCircumstanceRepository = personalCircumstanceRepository;
+        this.personalCircumstanceTransformer = personalCircumstanceTransformer;
+    }
+
+    public List<PersonalCircumstance> personalCircumstancesFor(Long offenderId) {
+        List<uk.gov.justice.digital.delius.jpa.standard.entity.PersonalCircumstance> personalCircumstances = personalCircumstanceRepository.findByOffenderId(offenderId);
+        return personalCircumstances
+                .stream()
+                .filter(personalCircumstance -> !convertToBoolean(personalCircumstance.getSoftDeleted()))
+                .sorted(Comparator.comparing(uk.gov.justice.digital.delius.jpa.standard.entity.PersonalCircumstance::getPersonalCircumstanceId).reversed())
```
andymarke: There is no sensible order, so just ordering be insertion reversed. Any one have any better ideas? Delius orders on the UI by `startDate`

src/main/resources/db/migration/V1_12__schema.sql
```diff
@@ -0,0 +1,142 @@
+CREATE TABLE R_CIRCUMSTANCE_SUB_TYPE
```
andymarke: Back fill some tables that we map but were missing for DEV

src/main/java/uk/gov/justice/digital/delius/data/api/PersonalCircumstance.java
llmikeyj: It's a toughie with anonymous tuples like this. My gut feeling would be to call it 'value' for want of a better vague term

src/main/java/uk/gov/justice/digital/delius/service/PersonalCircumstanceService.java
llmikeyj: Unless you understand the subject of 'more important' Personal Circumstances, I think that's about all we can do unless there's useful timestamps maybe

src/main/java/uk/gov/justice/digital/delius/data/api/PersonalCircumstance.java
```diff
@@ -0,0 +1,21 @@
...
+    private KeyValue subType;
+    private LocalDateTime startDate;
```
llmikeyj: Do these need to be date times? Or would date suffice?

src/main/java/uk/gov/justice/digital/delius/transformers/PersonalCircumstanceTransformer.java
```diff
@@ -0,0 +1,52 @@
...
+                .probationArea(Optional.ofNullable(personalCircumstance.getProbationArea()).map(this::probationAreaOf).orElse(null))
+                .endDate(personalCircumstance.getEndDate())
+                .startDate(personalCircumstance.getStartDate())
+                .offenderId(personalCircumstance.getOffenderId())
+                .notes(personalCircumstance.getNotes())
+                .evidenced(ynToBoolean(personalCircumstance.getEvidenced()))
+            .build();
+    }
+
+    private KeyValue probationAreaOf(ProbationArea probationArea) {
```
llmikeyj: This transform probably already exists in another transformer

andymarke: The key/value pair version is duplicated in a few transformers, perhaps we should add  a `public KeyValue keyValueOf(ProbabtionArea probationArea)` to the ProbationAreaTransformer` along side the existing `probationAreaOf` used by the ReferenceData service? Feels a bit OTT though

src/main/java/uk/gov/justice/digital/delius/data/api/PersonalCircumstance.java
andymarke: yeah, I found a precedence with `Nsi` model, which calls it `nsiType` and `nsiSubType` - so renamed to be consistent.

PR #18, opened 2018-11-05
Convictions

retroredge:


src/main/java/uk/gov/justice/digital/delius/controller/InstitutionalReportController.java
```diff
@@ -0,0 +1,118 @@
+package uk.gov.justice.digital.delius.controller;
```
llmikeyj: General comment, I thought we were going to surface specific types of institutional report, eg parole report?

src/main/java/uk/gov/justice/digital/delius/transformers/InstitutionalReportTransformer.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Optional;
+
+import static uk.gov.justice.digital.delius.transformers.TypesTransformer.convertToBoolean;
+import static uk.gov.justice.digital.delius.transformers.TypesTransformer.zeroOneToBoolean;
+
+@Component
+public class InstitutionalReportTransformer {
+
+    public InstitutionalReportTransformer() {
```
llmikeyj: Do we need this

src/main/java/uk/gov/justice/digital/delius/transformers/InstitutionalReportTransformer.java
```diff
@@ -0,0 +1,44 @@
...
+            .institutionalReportId(report.getInstitutionalReportId())
+            .offenderId(report.getOffenderId())
+            .conviction(convictionOf(report))
+            .build();
+    }
+
+    private Conviction convictionOf(uk.gov.justice.digital.delius.jpa.standard.entity.InstitutionalReport report) {
+
+        return Optional.ofNullable(report.getCustody())
+            .filter(custody -> !convertToBoolean(custody.getSoftDeleted()))
```
andymarke: Minor observation; I bet this can't happen in the system; e.g we have a non-deleted report associated with a deleted custody/disposal record. But the code does no harm so no need to change

retroredge: Hopefully it would never happen. This code protects us if it does. BCL code or DB triggers should keep all the soft deleted flags in sync across connected records

PR #22, opened 2018-12-11
Offender registrations

andymarke:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Team.java
```diff
@@ -38,8 +38,9 @@
     @JoinColumn(name = "LOCAL_DELIVERY_UNIT_ID")
     private LocalDeliveryUnit localDeliveryUnit;
 
-    @Column(name = "PROBATION_AREA_ID")
-    private Long probationAreaId;
+    @JoinColumn(name = "PROBATION_AREA_ID")
```
andymarke: For some reason all these columns are marked as `OneToOne` rather than `ManyToOne` so continued with same pattern for this entity

PR #25, opened 2018-12-20
Offender convictions

andymarke:


src/main/java/uk/gov/justice/digital/delius/data/api/InstitutionalReport.java
```diff
@@ -1,13 +1,20 @@
...
+
 @Data
 @Builder(toBuilder = true)
 public class InstitutionalReport {
     private Long institutionalReportId;
     private Long offenderId;
     private Conviction conviction;
-    private Sentence sentence;
+
+    @ApiModelProperty(notes = "Deprecated - Use conviction to access sentence")
```
andymarke: I will remove this property once New Tech is updated

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Event.java
```diff
@@ -23,18 +22,40 @@
...
     @Column(name = "CONVICTION_DATE")
     private LocalDate convictionDate;
 
+    @Column(name = "REFERRAL_DATE")
+    private LocalDate referralDate;
+
     @Column(name = "SOFT_DELETED")
     private Long softDeleted;
+
+    @OneToOne(mappedBy = "event")
```
andymarke: All these back references explain why PR is much bigger than expected; much of the original work NR did for retrieving sentences, offences etc from court/institutional report can change from additional service calls to navigating the Delius model.

src/main/java/uk/gov/justice/digital/delius/service/InstitutionalReportService.java
```diff
@@ -49,20 +45,7 @@ public InstitutionalReportService(InstitutionalReportRepository institutionalRep
...
-
-        return Optional.ofNullable(institutionalReport.getConviction())
-            .map(ignored -> institutionalReport.toBuilder()
-                .conviction(
-                    institutionalReport.getConviction().toBuilder()
-                        .offences(offenceService.eventOffencesFor(institutionalReport.getConviction().getConvictionId()))
-                        .build())
-                .build())
-            .orElseGet(() -> institutionalReport);
+                .map(institutionalReportTransformer::institutionalReportOf);
```
andymarke: No longer need additional service call; just navigate from *Event*

src/main/java/uk/gov/justice/digital/delius/service/OffenceService.java
```diff
@@ -42,24 +39,13 @@ public OffenceService(MainOffenceRepository mainOffenceRepository, AdditionalOff
...
     }
 
-    ImmutableList<Offence> eventOffencesFor(Long eventId) {
-        return combineMainAndAdditionalOffences(mainOffenceFor(eventId));
-    }
-
     private ImmutableList<Offence> combineMainAndAdditionalOffences(MainOffence mainOffence) {
         List<Offence> additionalOffences =
-            additionalOffenceTransformer.offencesOf(additionalOffenceRepository.findByEventId(mainOffence.getEventId()));
+            additionalOffenceTransformer.offencesOf(mainOffence.getEvent().getAdditionalOffences());
```
andymarke: No longer need additional service call; just navigate from *Event*

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+                .index(indexOf(event.getEventNumber()))
+                .offences(offencesOf(event))
+                .sentence(Optional.ofNullable(event.getDisposal()).map(this::sentenceOf).orElse(null))
+                .inBreach(zeroOneToBoolean(event.getInBreach()))
+                .latestCourtAppearanceOutcome(Optional.ofNullable(event.getCourtAppearances()).map(this::outcomeOf).orElse(null))
+                .build();
+    }
+
+    private Long indexOf(String eventNumber) {
+        // no DB constraint so protected ourselves
```
andymarke: Oddly *EVENT_NUMBER* always contains a number but stored as a String but without any Oracle constraints - so have to be cautious mapping it

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+                .filter(this::isNumber)
+                .map(text -> NumberUtils.parseNumber(text, Long.class))
+                .orElse(null);
+    }
+
+    private boolean isNumber(String text) {
+        return text.matches("\\d*");
+    }
+
+    private KeyValue outcomeOf(List<CourtAppearance> courtAppearances) {
```
andymarke: This is defined as the latest court appearance where the outcome is recorded; i.e. it would ignore a future court case which has no outcome yet

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+                .builder()
+                .description(outcome.getCodeDescription())
+                .code(outcome.getCodeValue())
+                .build();
+    }
+
+
+    private List<Offence> offencesOf(Event event) {
+        return ImmutableList.<Offence>builder()
+                .addAll(Optional.ofNullable(event.getMainOffence()).map(mainOffence -> ImmutableList.of(mainOffenceTransformer.offenceOf(mainOffence))).orElse(ImmutableList.of()) )
```
andymarke: Database would in theory allow an event without an offence, but in practice UI makes this impossible - either way I protect myself from bad data with the Null check

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+                .addAll(additionalOffenceTransformer.offencesOf(event.getAdditionalOffences()))
+                .build();
+    }
+
+    private Sentence sentenceOf(Disposal disposal) {
```
andymarke: Moved for Institutional Report transformer

src/main/java/uk/gov/justice/digital/delius/transformers/CourtAppearanceTransformer.java
```diff
@@ -29,7 +29,7 @@ public CourtAppearance courtAppearanceOf(uk.gov.justice.digital.delius.jpa.stand
             .crownCourtCalendarNumber(courtAppearance.getCrownCourtCalendarNumber())
             .bailConditions(courtAppearance.getBailConditions())
             .courtNotes(courtAppearance.getCourtNotes())
-            .eventId(courtAppearance.getEventId())
+            .eventId(courtAppearance.getEvent().getEventId())
```
andymarke: Not sure why we return the eventId - I think it is due to New Tech making additional calls to get Offences...

src/test/java/uk/gov/justice/digital/delius/service/CourtAppearanceServiceTest.java
```diff
@@ -52,7 +44,11 @@ public void setUp() {
                         .softDeleted(1L)
                         .appearanceDate(LocalDateTime.now())
                         .offenderId(1L)
-                        .eventId(50L)
```
andymarke: All these tests change to reflect the Model changes and removal of service calls...

src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
```diff
@@ -10,7 +10,12 @@
 @Builder(toBuilder = true)
 public class Conviction {
     private Long convictionId;
+    private Long index;
```
llmikeyj: What does index mean?

src/main/java/uk/gov/justice/digital/delius/data/api/InstitutionalReport.java
llmikeyj: ok

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Event.java
llmikeyj: ok

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/MainOffenceRepository.java
```diff
@@ -7,6 +7,4 @@
 
 public interface MainOffenceRepository extends JpaRepository<MainOffence, Long> {
     List<MainOffence> findByOffenderId(Long offenderId);
-
```
llmikeyj: Is this safe? What about the offender main offence for a prior conviction?

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
llmikeyj: Does it really matter? What if you just surface it as a String?

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+        return Sentence.builder()
+                .defaultLength(disposal.getLength())
+                .effectiveLength(disposal.getEffectiveLength())
+                .lengthInDays(disposal.getLengthInDays())
+                .originalLength(disposal.getEntryLength())
+                .originalLengthUnits(Optional.ofNullable(disposal.getEntryLengthUnits())
+                        .map(StandardReference::getCodeDescription)
+                        .orElse(null))
+                .secondLength(disposal.getLength2())
```
llmikeyj: Did we not find a better way of describing "second length"? Sure I badgered Nick R about if for an earlier PR

src/main/java/uk/gov/justice/digital/delius/transformers/CourtAppearanceTransformer.java
llmikeyj: Leave it in for now I think, it allows for some wriggle room when traversing the model for other use cases.

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
andymarke: No, and unfortunately the Delius BA essentially described it as multi use - some times just the second half of a sentence (3 years in prison and 4 years on licence) and sometimes for suspended period for suspended sentence. In other words multi-use so we can't find a name that describes it better

src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
andymarke: Good question; in Delius they use it as an identifier for an event in the UI. e.g the second Event added to Delius is known as the "event number 2". So it could be called the `offenderConvictionId` but I feel this might confuse any client in thinking it is some sort of unique id. On database they call it EVENT_NUMBER so I think index might be the best term, maybe `offenderIndex`

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
andymarke: Yep you are quite right. Changed it back to String

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
llmikeyj: Actually, I really don't like this. Current understanding is that the field in Delius has 2 meanings, one as the tariff length for a life sentence, and the other the suspended sentence period for suspended sentences. I would propose, then, taking `secondLength` and associated fields out of the API response, and replacing them with either `lifeSentenceTariff` or `suspendedSentencePeriod`

andymarke: The problem is we don't know if it is a `suspendedSentencePeriod ` or a `licenceLength` or a `lifeSentenceTariff `. We just don't have enough context for this given we have to model a legacy database. So at best we would have to fraudulently populate all those fields even though length2 for a particular sentence might mean to the probation service only one of those things. I'm not sure we can change this until we have control on the data written and we can model that correctly. Right now it is literally the thing as described 'a second sentence length'

src/main/java/uk/gov/justice/digital/delius/transformers/CourtAppearanceTransformer.java
andymarke: sure

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/MainOffenceRepository.java
andymarke: Yes, since a main offence can only exist in the  context of an event and event only exits when they have been convicted (though they can be created before they are sentenced). So finding by event Id really means `give me the offence associated with this conviction`

llmikeyj: But is this not 'find all my main offences including those related to the current event'?

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
llmikeyj: What an utter shambles. It's just a skip whose contents are only ever referenced by the front end, that's what we're saying?

PR #26, opened 2019-01-17
316 added exclusion restriction message

andymarke:


andymarke:
Two changes in this PR
* For userAccess offender checks now return the message indicating why the user has no access and who to contact
* A correction to offender manager allocation reason - I was original getting this from the offender transfer table, but missed that the reason is also held on the offender manager record - which is far easier to get. It is also populated for initial allocation where as the transfer table isn't

src/test/java/uk/gov/justice/digital/delius/service/UserServiceTest.java
```diff
@@ -0,0 +1,280 @@
...
+    public void exclusionListNotCheckedWhenNotExcludedForAnyone() {
+        userService.accessLimitationOf("Micky", OffenderDetail
+                .builder()
+                .currentExclusion(false)
+                .currentRestriction(false)
+                .exclusionMessage("How dare you - you are excluded")
+                .restrictionMessage("How dare you - you are restricted")
+                .build());
+
+        verify(userRepositoryWrapper, never()).getUser(Mockito.any());
```
llmikeyj: Not clear to me what the value of this test is?

PR #27, opened 2019-01-25
323 offender disability

andymarke:


src/main/java/uk/gov/justice/digital/delius/data/api/Disability.java
```diff
@@ -0,0 +1,16 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+
+import java.time.LocalDate;
+
+@Data
+@Builder
+public class Disability {
+    private Long disabilityId;
```
llmikeyj: Why do we need to surface the Id through the API? Is this the PK or something else?

src/main/java/uk/gov/justice/digital/delius/data/api/Disability.java
```diff
@@ -0,0 +1,16 @@
...
+    private KeyValue type;
```
llmikeyj: What does 'type' mean? Is there a better description?

src/test/java/uk/gov/justice/digital/delius/transformers/OffenderTransformerTest.java
```diff
@@ -33,6 +40,90 @@ public void offenderManagerAllocationReasonNullWhenOffenderTransferAbsent() {
 
     }
 
+    @Test
+    public void disabilitiesCopiedWithLatestFirst() {
+        assertThat(idsOf(offenderTransformer.fullOffenderOf(anOffender()
```
llmikeyj: You don't really need idsOf(...), you can use something like assertThat(offenderTransformer.fullOffenderOf(..)).extracting("disabilityId").containsExactly(...)

src/main/java/uk/gov/justice/digital/delius/data/api/Disability.java
andymarke: Yeah, primary key for a disability - generally clients need something that can be used to distinguish between the disabilities - though are unlikely to use in subsequent API calls until we have update/delete functionality

src/main/java/uk/gov/justice/digital/delius/data/api/Disability.java
andymarke: Renamed to disabilityType

llmikeyj:
lgtm!

PR #32, opened 2019-05-31
Licences responsible officer

timharrison-moj:
This branch contains the full HDC licences changes to provide managed offenders for a staff officer and responsible officers for an offender. Both views are generic and can be limited to currently assigned officers with the ?current=true query parameter.


src/main/java/uk/gov/justice/digital/delius/controller/OffenderController.java
```diff
@@ -372,6 +366,22 @@ private HttpStatus responseCodeOf(AccessLimitation accessLimitation) {
...
+    @ApiResponses(value = {
+                      @ApiResponse(code = 200, message = "A list of responsible officers for an offender", response = ResponsibleOfficer.class, responseContainer = "List"),
+                      @ApiResponse(code = 401, message = "Request is missing Authorization header (no JWT)"),
+                      @ApiResponse(code = 403, message = "The requesting user was restricted from access", response = AccessLimitation.class),
+                      @ApiResponse(code = 404, message = "The requested offender was not found.")
+            })
+    @JwtValidation
+    public ResponseEntity<List<ResponsibleOfficer>> getResponsibleOfficersByNomsNumber(final @RequestHeader HttpHeaders httpHeaders,
+                                                                                       final @PathVariable("nomsNumber") String nomsNumber,
+                                                                                       final @RequestParam(name="current", required=false, defaultValue="false") String current) {
```
llmikeyj: Can we model this as a boolean?

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
```diff
@@ -0,0 +1,71 @@
...
+    })
+    @JwtValidation
+    public ResponseEntity<List<ManagedOffender>> getManagedOffendersByStaffCode(final @RequestHeader HttpHeaders httpHeaders,
+                                                                                final @PathVariable("staffCode") String staffCode,
+                                                                                final @RequestParam(name="current", required=false, defaultValue="false") String current) {
+        return staffService.getManagedOffendersByStaffCode(staffCode, current)
+                .map(managedOffenders -> new ResponseEntity<>(managedOffenders ,OK))
+                .orElse(new ResponseEntity<>(NOT_FOUND));
+
+        // TODO: Check this - only NOT_FOUND if the staffCode was not an existing officer
```
llmikeyj: What's the outcome for this TODO?

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.OK;
+
+@RestController
+@Slf4j
+@Api(tags = "Staff")
+public class StaffController
+{
+    private StaffService staffService;
+    private Jwt jwt;
```
llmikeyj: I don't think this is needed

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.OK;
+
+@RestController
+@Slf4j
+@Api(tags = "Staff")
+public class StaffController
+{
+    private StaffService staffService;
```
llmikeyj: Prefer final

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @ExceptionHandler(HttpServerErrorException.class)
+    public ResponseEntity<String> restServerError(HttpServerErrorException e) {
+        return new ResponseEntity<>(e.getMessage(), e.getStatusCode());
+    }
+
+    @ExceptionHandler(NoSuchUserException.class)
+    public ResponseEntity<String> noSuchUser(NoSuchUserException e) {
+        return new ResponseEntity<>(e.getMessage(), NOT_FOUND);
+    }
```
llmikeyj: May want to consider moving these exception handlers into a new ControllerAdvice, alongside those in the OffenderController? Now that we have a new controller context?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
```diff
@@ -195,8 +188,10 @@
     @JoinColumn(name = "PARTITION_AREA_ID")
     private PartitionArea partitionArea;
 
-    @JoinColumn(name = "OFFENDER_ID")
-    @OneToMany
+    @OneToMany(mappedBy = "offenderId", fetch = FetchType.LAZY)
```
llmikeyj: LAZY is default fetch

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
```diff
@@ -316,4 +305,147 @@ private Human humanOf(String forename, String forename2, String surname) {
     }
 
 
+    public List<uk.gov.justice.digital.delius.data.api.ResponsibleOfficer> responsibleOfficersOf(Offender offender, boolean current) {
+
+        ArrayList<ResponsibleOfficer> responsibleOfficers = new ArrayList<>();
+
+        if (offender.getOffenderManagers() != null && !offender.getOffenderManagers().isEmpty()) {
+
+            for (uk.gov.justice.digital.delius.jpa.standard.entity.OffenderManager om : offender.getOffenderManagers()) {
```
llmikeyj: Can we do this in a Streaming style please?

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
```diff
@@ -316,4 +305,147 @@ private Human humanOf(String forename, String forename2, String surname) {
...
+
+                boolean isCurrentRo   = (om.getResponsibleOfficer() != null && om.getResponsibleOfficer().getEndDate() == null);
+                boolean isCurrentOm   = (om.getActiveFlag().intValue() == 1 && om.getEndDate() == null);
```
llmikeyj: Consider pulling these out into functions and testing independently?

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
```diff
@@ -316,4 +305,147 @@ private Human humanOf(String forename, String forename2, String surname) {
...
+
+        if (current && !responsibleOfficers.isEmpty()) {
+            // Filter the list to only those officers who hold a currently active role for this offender
+            return responsibleOfficers
+                    .stream()
+                    .filter(ro -> (ro.isCurrentOm() || ro.isCurrentPom() || ro.isCurrentRo()))
+                    .collect(Collectors.toList());
+        }
+
+        return responsibleOfficers;
```
llmikeyj: Feels quite a chunky method this

src/test/java/uk/gov/justice/digital/delius/StaffRepositoryTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import javax.transaction.Transactional;
+import java.util.Optional;
+
+/*
+ * Uncomment this test and run manually to ensure that the data is loading
+ * and a Staff record can be found by ID.
+ */
+
+@Ignore
```
llmikeyj: What we doing with this?

src/test/java/uk/gov/justice/digital/delius/transformers/OffenderTransformerTest.java
```diff
@@ -50,7 +54,7 @@ public void disabilitiesCopiedWithLatestFirst() {
                                 .disabilityType(StandardReference
                                         .builder()
                                         .codeValue("A")
-                                        .codeDescription("Metal health")
```
llmikeyj: \m/

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
timharrison-moj: Removed this - the tests already confirmed that and unknown staff code passed in will result in a 404.

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
timharrison-moj: You are right. I think this has been added, then functionality removed or moved elsewhere and its no longer used. I will remove it.

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
timharrison-moj: Agreed, and done.

src/main/java/uk/gov/justice/digital/delius/controller/StaffController.java
timharrison-moj: The only @ControllerAdvice I've seen in the application is for Jwt validation. For offenders, the exception handlers are as they are in the StaffController with local @ExceptionHandler methods, but yes, creating shared advice would save the controllers from duplicating these handlers so look at moving these.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
timharrison-moj: You are right - I've altered this one to remove the specific fetch type. For info, for @ManyToOne relationships it default to FetchType.EAGER.

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
timharrison-moj: We certainly can. Working on it.

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
timharrison-moj: While looking at the streaming I will also see if I can break this down - I agree, it is chunky with two blocks of quite similar code.

src/test/java/uk/gov/justice/digital/delius/StaffRepositoryTest.java
timharrison-moj: Its a useful test, though does not assert anything at present. I used it interactively to check that the StaffRepository methods could find staff members by methods, much like a similar offenderRepository test does. I will remove the @Ignore and assert the expected value.

src/test/java/uk/gov/justice/digital/delius/transformers/OffenderTransformerTest.java
timharrison-moj: The base data actually sets this up and then checks for it. I had a note to check this with you!

src/main/java/uk/gov/justice/digital/delius/controller/OffenderController.java
timharrison-moj: Yes, done this.

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
timharrison-moj: Done.

src/main/java/uk/gov/justice/digital/delius/data/api/ManagedOffender.java
```diff
@@ -0,0 +1,32 @@
...
+    @ApiModelProperty(required = true)
+    private Long offenderId;
+    @ApiModelProperty(required = true)
+    private String nomsNumber;
+    @ApiModelProperty(required = true)
+    private String crnNumber;
+    @ApiModelProperty(required = true)
+    private String offenderSurname;
+    @ApiModelProperty(required = true)
+    private boolean isCurrentRo;
```
andymarke: Minor; might be worth expanding this `isCurrentRo` to `isCurrentResponsibleOfficer` they are a bit cryptic to a newbie

src/main/java/uk/gov/justice/digital/delius/data/api/ResponsibleOfficer.java
```diff
@@ -0,0 +1,42 @@
...
+    @ApiModelProperty
+    private Long offenderManagerId;
+    @ApiModelProperty
+    private Long prisonOffenderManagerId;
+    @ApiModelProperty(required = true)
+    private String staffCode;
+    @ApiModelProperty(required = true)
+    private String surname;
+    private String forenames;
+    private String providerTeamCode;
```
andymarke: Minor: maybe model these with the Team, Area, ProbationArea classes?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
```diff
@@ -0,0 +1,64 @@
...
+@Table(name = "PRISON_OFFENDER_MANAGER")
+public class PrisonOffenderManager {
+
+    @Id
+    @Column(name = "PRISON_OFFENDER_MANAGER_ID")
+    private Long prisonOffenderManagerId;
+
+    @JoinColumn(name = "ALLOCATION_TEAM_ID")
+    @OneToOne
+    private Team team;
```
andymarke: Double check this shouldn't point to AllTeam - I am only just getting my head round the differences between Team and AllTeam and I'm getting the impression AllTeam should be used but I think we need to do a bit more digging..

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @JoinColumn(name = "PROBATION_AREA_ID")
+    @OneToOne
+    private ProbationArea probationArea;
+
+    @Column(name = "ACTIVE_FLAG")
+    private Long activeFlag;
+
+    @Column(name = "ALLOCATION_DATE")
+    private Timestamp allocationDate;
```
andymarke: I think the pattern is to use LocaDate and LocalDateTime in our entities

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
```diff
@@ -0,0 +1,64 @@
...
+    @JoinColumn(name = "ALLOCATION_TEAM_ID")
+    @OneToOne
+    private Team team;
+
+    @JoinColumn(name = "ALLOCATION_STAFF_ID")
+    @OneToOne
+    private Staff staff;
+
+    @Column(name = "OFFENDER_ID")
+    private Long offenderId;
```
andymarke: Do you need this and `managedOffender` ?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ResponsibleOfficer.java
```diff
@@ -0,0 +1,48 @@
...
+    @Column(name = "RESPONSIBLE_OFFICER_ID")
+    private Long responsibleOfficerId;
+
+    /*
+     Currently - the responsible officer can be either the OM or the POM.
+     Until the POM is populated with realistic data by the OMiC programme the RESPONSIBLE_OFFICER will not contain accurate data for Prison staff.
+     JPA links to the OFFENDER entity should be via the RESPONSIBLE_OFFICER table but are currently via OFFENDER_MANAGER and PRISON_OFFENDER_MANAGER entiries.
+     When real POMs are present it will make sense to remove existing links and add a single link via RESPONSIBLE_OFFICER to the OFFENDERS @manyToOne.
+     */
+
```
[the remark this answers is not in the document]
andymarke: Might be worth we get together with @llmikeyj  - doesn't feel we should need to change the code due to the data

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Staff.java
```diff
@@ -31,4 +28,14 @@
...
+    @Column(name = "OFFICER_CODE")
+    private String officerCode;
+
+    @OneToMany
+    @JoinColumn(name = "ALLOCATION_STAFF_ID")
+    List<OffenderManager> offenderManagers;
+
+    @OneToMany
+    @JoinColumn(name = "ALLOCATION_STAFF_ID")
+    List<PrisonOffenderManager> prisonOffenderManagers;
```
andymarke: Same as above

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
```diff
@@ -315,5 +304,169 @@ private Human humanOf(String forename, String forename2, String surname) {
...
+                    .filter(mo -> (mo.isCurrentOm() || mo.isCurrentPom() || mo.isCurrentRo()))
+                    .collect(Collectors.toList());
+        }
+
+        return managedOffenders;
+    }
+
+
+    private boolean isCurrentRo(uk.gov.justice.digital.delius.jpa.standard.entity.ResponsibleOfficer ro) {
+        boolean result = false;
```
andymarke: Minor: Maybe we could do `return Optional.ofNullable(ro).filter(not(Responsible:Officer::getEndDate)).isPresent()` for readability

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ResponsibleOfficer.java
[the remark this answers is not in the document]
timharrison-moj: After talking with the OMiC team - they intend to make changes in this area in order to set the responsible officer earlier in the cycle whilst the offender is still in prison. At present those changes are not defined or made, so I left this comment in just to note the intended actions. The code will work now in all environments as it is.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Staff.java
```diff
@@ -31,4 +28,14 @@
     @Column(name = "FORENAME2")
     private String forname2;
 
+    @Column(name = "OFFICER_CODE")
+    private String officerCode;
+
+    @OneToMany
+    @JoinColumn(name = "ALLOCATION_STAFF_ID")
```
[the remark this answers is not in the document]
timharrison-moj: As we discussed earlier - I checked several areas and identified the foreign keys being used, and the allocation staff id was linked back to the STAFF entity to identify the person being allocated. I think we should trial it as is, and then make any required changes if its not working as intended? The two changes made are read-only, and no clients currently use them, so its low risk.

src/main/java/uk/gov/justice/digital/delius/data/api/ResponsibleOfficer.java
timharrison-moj: Possibly, but the requirements for HDC were fairly narrow - what I've done is to try to make the response objects simple but flexible enough for other use-cases besides HDC. For example, the endpoints can be limited to currently assigned officers using the ?current=true parameter which is what HDC require, but omitting that will provide the previous allocations of ROs and OMs too.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
timharrison-moj: Good point. I was confused by the AllTeam relationships here. I am thinking that it might be better to send this out (low risk) and let HDC check whether the data they see matches up with the data they currently get from MI extracts. I can then investigate any differences and correct the relationships if required?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
timharrison-moj: Fair point - I will look to change that shortly. The other APIs (elite2 etc) all use LocalDateTime and LocalDate, so it does stand out as an oddity.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/PrisonOffenderManager.java
timharrison-moj: Yes I think so. I will have a check shortly to see where it is used.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Staff.java
timharrison-moj: Response above too.

src/main/java/uk/gov/justice/digital/delius/transformers/OffenderTransformer.java
timharrison-moj: I will take a look again and see which reads best. Thanks.

PR #36, opened 2019-06-25
DT-8: Add new ping endpoint and document

SimonMitchellMOJ:


src/test/java/uk/gov/justice/digital/delius/info/PingEndpointTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    public void healthEndpointIsUnauthorised() {
+        String responseBody = given()
+                .when()
+                .get("/ping")
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.OK.value())
```
andrewrlee: can we check that the content type is plain/text?

petergphillips: I spoke to @SimonMitchellMOJ about this and agreed to pick it up in his next PR :grin:

andymarke:
Also increment version number in build.gradle

PR #38, opened 2019-08-02
628 ldap authentication

andymarke:


src/main/java/uk/gov/justice/digital/delius/controller/AuthenticationController.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @RequestMapping(value = "/users/{username}/details", method = RequestMethod.GET)
+    public ResponseEntity<UserDetails> findUser(final @PathVariable("username") String username) {
+        return userService.getUserDetails(username).map(userDetails -> new ResponseEntity<>(userDetails, OK))
+                .orElse(new ResponseEntity<>(NOT_FOUND));
+
+    }
+
+    @RequestMapping(value = "/users/{username}/password", method = RequestMethod.POST)
```
timharrison-moj: Should there be a safeguard against any call to the API from being able to lock/unlock/change passwords for any user? just wondering if the API should be protected to specific users/roles?  I know its currently in a locked down environment, but without authorisation on the caller could any client could use these endpoints?

src/main/java/uk/gov/justice/digital/delius/ldap/repository/entity/NDeliusUser.java
```diff
@@ -0,0 +1,29 @@
...
+@AllArgsConstructor
+public class NDeliusUser {
+    @Id
+    private Name dn;
+    private String cn;
+    private String sn;
+    private String mail;
+    private String givenname;
+    private String orclActiveEndDate;
+    @Transient
```
timharrison-moj: Is this @Transient because roles come from LDAP not the database? So a persist/save never tries to write roles. Does hte database not contain roles? Just a question for my understanding.

src/main/java/uk/gov/justice/digital/delius/controller/AuthenticationController.java
```diff
@@ -0,0 +1,65 @@
...
+public class AuthenticationController {
+
+    private final UserService userService;
+
+    public AuthenticationController(UserService userService) {
+        this.userService = userService;
+    }
+
+    @RequestMapping("/authenticate")
+    public ResponseEntity authenticate(final @RequestParam String username, @RequestParam String password) {
```
timharrison-moj: Is this called using HTTPS or HTTP? Just thinking that the password in the body will not be encrypted in flight over HTTP, so visible in logs (and packet sniffers).  Do all clients use the Tolomy HTTPS gateway to access the community API? Its fine if they do.

src/main/java/uk/gov/justice/digital/delius/controller/AuthenticationController.java
andymarke: Yes they could right now. And until the API is protected this is exactly how it works now. Any client can masquerade as any user without a password. There is essentially no security other than this is in UKCloud. I did think about adding some sort of protection just for this APIs but I would be wasting my time since I can still access all data anyway with no credentials.

My gut feeling is we will eventually protect these APIs with a service to service token (issued by AUTH) and everything else will be the user's token (also issued by AUTH). But by definition we have no AUTH server right now so we can't.

One final thing to note; in live there is no password in LDAP it is in active directory so this is all post-aws migration work so we can't do any damage in live. The APIs essentially only work in test environments. Hence it is still a spike that can have no side-affect in live even if someone to ssh on to a UKCloud box or got through your NDH tunnel.

src/main/java/uk/gov/justice/digital/delius/ldap/repository/entity/NDeliusUser.java
andymarke: They do but they are not directly in the same LDAP tree, you have to do a separate ldap search to get them . They are set as aliases to other objects. So we load the NDeliusUser and then fix up the roles using a second call (Spring ldapTemplate would fail if it wasn't transient)

src/main/java/uk/gov/justice/digital/delius/controller/AuthenticationController.java
andymarke: All http within UKCloud I'm afraid. Along with all sensitive probation data. So from client to auth app it would be https, then from MoJ AWS Cloud platform to any UKCLoud gateway would be https but once in side UKCloud it would be plain text. The same as Delius now

andymarke: Actually even from the browser client to Delius in probation is plain http since they are on a "secure"  VPN . So believe it or not Delius uses http in live. Presumably this will all change when it is delius is migrated to AWS.

andymarke: Although in pre-prod it is SSL but with and invalid certificate. But prod is plain http

PR #39, opened 2019-08-02
628 Added feature flag to give hint that this isn't being used live yet

andymarke:


build.gradle
```diff
@@ -28,6 +28,11 @@ springBoot {
     buildInfo()
 }
 
+test {
+    testLogging.showStandardStreams = true
```
andymarke: We are getting a lot of builds hanging so stuck this in to find out at what point it hangs.

timharrison-moj: I think there is one test - around documents - which takes a really long time, and even fails 50% of the time for me. Circle usually succeeds. Might even be worth taking this test out for now?

andymarke: I take a look. Thanks

PR #40, opened 2019-08-05
Attempt to fix hanging test by using the "not found"  document

andymarke:


src/test/java/uk/gov/justice/digital/delius/DeliusOffenderAPITest.java
```diff
@@ -465,7 +465,7 @@ public void getOffenderDocumentDetailsForKnownOffenderCRNButUnknownDocumentGives
         given()
                 .header("Authorization", aValidToken())
                 .when()
-                .get("/offenders/crn/D002384/documents/" + UUID.randomUUID().toString() + "/detail")
+                .get("/offenders/crn/D002384/documents/fa63c379-8b31-4e36-a152-2a57dfe251c5/detail")
```
andymarke: This id has an associated WireMock not found response configured. Not sure why generating a random request would sometimes hand wiremock but this appears to fix the problem

PR #50, opened 2019-09-25
Dt 214 graceful when document link missing

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/AlfrescoService.java
```diff
@@ -56,21 +53,11 @@ public SearchResult listDocuments(String crn) {
     }
 
     public ResponseEntity<Resource> getDocument(String documentId, String crn) {
+        val maybeDocumentMeta = getDocumentDetail(documentId, crn);
 
-        // Check crn exists
```
andymarke: Previous this was retrieving *all* documents for an offender and finding the one it wants rather than simply asking Alfresco for the specific document (and then comparing crn)

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/AddressAssessmentDocumentRepository.java
```diff
@@ -1,10 +1,13 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 import uk.gov.justice.digital.delius.jpa.standard.entity.AddressAssessmentDocument;
 
 import java.util.List;
 
 public interface AddressAssessmentDocumentRepository extends JpaRepository<AddressAssessmentDocument, Long> {
-    List<AddressAssessmentDocument> findByOffenderId(Long offenderId);
+    @Query("select document from AddressAssessmentDocument document, AddressAssessment entity where document.addressAssessment = entity and document.offenderId = :offenderId")
+    List<AddressAssessmentDocument> findByOffenderId(@Param("offenderId") Long offenderId);
```
andymarke: Query now limits so parent is not found if child is not found by joining to the table table. Required since Delius has bad data with documents pointing to entities that no longer exist

PR #51, opened 2019-09-27
DT-3: updated to Springboot 2 and add app insights

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/delius/config/AlfrescoConfig.java
```diff
@@ -14,7 +19,11 @@ public RestTemplate alfrescoRestTemplate(RestTemplateBuilder restTemplateBuilder
                                              @Value("${alfresco.baseUrl}") String baseUrl) {
         return restTemplateBuilder
                 .rootUri(baseUrl)
+                .additionalInterceptors(getRequestInterceptors())
```
petergphillips: in future can just add the interceptors here since the method takes varargs

src/test/java/uk/gov/justice/digital/delius/transformers/MainOffenceTransformerTest.java
```diff
@@ -16,6 +16,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.when;
 
+;
```
andymarke: minor typo?

src/main/resources/application.yml
```diff
@@ -1,47 +1,73 @@
 spring:
-  datasource:
```
petergphillips: yml files should be formatted with two spaces rather than four, please could you convert back

src/main/resources/application.yml
```diff
@@ -1,47 +1,73 @@
...
-    users:
-      base: "cn=Users,dc=moj,dc=com"
\ No newline at end of file
+    ldap:
+        users:
+            base: "cn=Users,dc=moj,dc=com"
+
+graceful:
+    shutdown:
+        enabled: true
```
petergphillips: can't see this set in the `dev` version?

src/test/java/uk/gov/justice/digital/delius/transformers/MainOffenceTransformerTest.java
petergphillips: strange addition to this file of `;`

build.gradle
```diff
@@ -1,90 +1,98 @@
...
 
-sourceCompatibility = 1.8
+def today = Instant.now()
+group = 'uk.gov.justice.digital.delius'
 
-// Apply application plugin for easy application start
-apply plugin: 'application'
+def todayDate = LocalDate.now()
+def formatter = DateTimeFormatter.ofPattern('yyyy-MM-dd')
+version todayDate.format(formatter)
```
andymarke: Are we missing build number on here? If we have two builds on same day would they not have the same version label? Looks jar name in Circle only has date is that correct?

petergphillips: Agreed, normally doesn't matter since we build docker containers instead.
Worth therefore grabbing the spring boot stuff and moving it up higher i.e. something like
```groovy
def todaysDate = LocalDate.now().format('yyyy-MM-dd')
version = System.getenv('CI') ? "${todaysDate}.${System.getenv('CIRCLE_BUILD_NUM')}" : todaysDate
```
and then reference `version` lower down instead

PR #59, opened 2019-10-17
DT-308 fix NPE when NSI is associated with offender rather than event

andymarke:


andymarke:
NPE caused by assuming Non-Statutory Interventions not always associated with an Event (conviction) but can be added at Offender level

src/test/java/uk/gov/justice/digital/delius/OffenderDocumentsAPITest.java
```diff
@@ -40,12 +39,6 @@
     @Autowired
     private ObjectMapper objectMapper;
 
-    @MockBean
```
andymarke: Convert test to integration from mocked test

PR #60, opened 2019-10-17
DT-309: Switch to java agent w3c tracing

petergphillips:


Dockerfile
```diff
@@ -1,6 +1,22 @@
-FROM java
-MAINTAINER Mike Jackson <michael.jackson@digital.justice.gov.uk>
+FROM openjdk:11-slim
```
petergphillips: this Dockerfile isn't used by the build process, so upgraded

build.gradle
```diff
@@ -61,25 +66,28 @@ dependencyManagement {
 dependencies {
     implementation files('lib/ojdbc8-12.2.0.1.jar')
 
-    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.9.10'
```
petergphillips: should really have pinned the same version as spring boot 2.1.9.RELEASE

PR #68, opened 2019-10-25
DT-247 Send IAPS notification of event change

andymarke:
For certain changes a notification is sent to a separate system that manages interventions called IAPS (Interim Accredited Programme System). This is handled by Delius by updating a record

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/SpgNotificationHelperRepositoryImpl.java
+++ b/src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/SpgNotificationHelperRepositoryImpl.java
@@ -1,16 +1,18 @@
 package uk.gov.justice.digital.delius.jpa.standard.repository;
 
+import lombok.val;
+import org.hibernate.Session;
+import org.hibernate.jdbc.ReturningWork;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/SpgNotificationHelperRepositoryImpl.java
```diff
@@ -21,20 +23,24 @@
 
     @Override
     public List<ProbationArea> getInterestedCRCs(String offenderId) {
```
andymarke: So  I have re-written old school. For reasons I don't understand this has started failing in Oracle (and we have no tests for Oracle). Not sure if it is related to Spring upgrades or what.
So a lot of discussion on this see for instance https://stackoverflow.com/questions/48847731/oracle-stored-procedure-java-sql-sqlexception-invalid-column-type-2012-ref-cu

Fundamentally `ParameterMode.REF_CURSOR` and how that is resolved needs to be `oracle.jdbc.OracleTypes.CURSOR` which I can only do with a callable statement.

Should be able to improve using JDBCTemplate etc - for now this seems to work.

PR #76, opened 2019-11-11
Dt 365 Move LDAP auth to /secure

Mjwillis:


src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
```diff
@@ -0,0 +1,118 @@
...
+    @ApiOperation(
+            value = "Authenticate a username and password against Delius Identity (LDAP)",
+            authorizations = {@Authorization("ROLE_AUTH_DELIUS_LDAP")})
+    @ApiResponses(
+            value = {
+                    @ApiResponse(code = 200, message = "OK"),
+                    @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+                    @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class)
+            })
+    @GetMapping("/authenticate")
```
Mjwillis: Should this be a POST instead?

petergphillips: Definitely.  If you have it as a `GET` then the username and password will appear in app insights

src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
```diff
@@ -0,0 +1,118 @@
...
+    @ApiResponses(
+            value = {
+                    @ApiResponse(code = 200, message = "Password Changed"),
+                    @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+                    @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
+                    @ApiResponse(code = 404, message = "Not Found", response = ErrorResponse.class)
+            })
+    @RequestMapping(value = "/users/{username}/password", method = RequestMethod.POST)
+    public ResponseEntity changePassword(@ApiParam(name = "username", value = "LDAP username", example = "TESTUSERNPS", required = true) final @PathVariable("username") String username,
+                                         @ApiParam(name = "password", value = "LDAP new password", example = "newpassword123", required = true) final @RequestParam("password") String password) {
```
petergphillips: same issue here with password as a request parameter

Mjwillis: Agreed was confused why it was done that way

PR #79, opened 2019-11-13
DT-362: added endpoint to expose Delius update case note endpoint

SimonMitchellMOJ:


src/main/java/uk/gov/justice/digital/delius/config/AlfrescoConfig.java
```diff
@@ -9,12 +9,22 @@
...
-    @Bean
+    @Bean("alfrescoRestTemplate")
     public RestTemplate alfrescoRestTemplate(RestTemplateBuilder restTemplateBuilder,
                                              @Value("${alfresco.baseUrl}") String baseUrl) {
         return restTemplateBuilder
                 .rootUri(baseUrl)
                 .build();
     }
 
+    @Bean("alfrescoRestTemplateWithAuth")
```
petergphillips: alfresco with auth :confused:

src/main/resources/application.yml
```diff
@@ -70,6 +70,9 @@ delius:
   ldap:
     users:
       base: "cn=Users,dc=moj,dc=com"
+  baseurl: 'http://localhost:8088/alfresco/s/noms-spg'
+  username: 'A'
+  password: 'A'
```
petergphillips: don't think we should have defaults for this in the application.yml

src/main/java/uk/gov/justice/digital/delius/service/CaseNoteService.java
```diff
@@ -0,0 +1,22 @@
...
+public class CaseNoteService {
+
+    private final RestTemplate restTemplate;
+
+    @Autowired
+    public CaseNoteService(@Qualifier("alfrescoRestTemplateWithAuth") RestTemplate restTemplate) {
+        this.restTemplate = restTemplate;
+    }
+
+    public void upsertCaseNotesToDelius(final String nomisId, final Long caseNotesId, final String caseNote) {
```
Mjwillis: Couldn't we model the object rather than a JSON string?

petergphillips: is there any point?  We're just acting as a proxy here so don't really care what the data is...

src/main/resources/application.yml
petergphillips: changed my mind based on the `dgoss` stuff running.  When that gets removed then we can remove this

src/main/java/uk/gov/justice/digital/delius/service/CaseNoteService.java
andymarke: Agree with Peter

PR #87, opened 2019-11-19
Ldap testing improvements

andymarke:


src/main/resources/application.yml
```diff
@@ -69,7 +69,7 @@ logging:
 delius:
   ldap:
     users:
-      base: "cn=Users,dc=moj,dc=com"
+      base: "ou=Users,dc=moj,dc=com"
```
andymarke: This is now the default in all environments post AWS migration

src/main/resources/embedded_ldap.properties
```diff
@@ -1,6 +1,5 @@
-spring.ldap.embedded.base-dn=dc=memorynotfound,dc=com
+spring.ldap.embedded.base-dn=dc=moj,dc=com
```
andymarke: Make in-memory LDAP same base as default real LDAP removing need for extra properties during testing etc

src/main/resources/embedded_ldap.properties
```diff
@@ -1,6 +1,5 @@
...
 spring.ldap.embedded.credential.username=uid=admin
 spring.ldap.embedded.credential.password=secret
 spring.ldap.embedded.ldif=classpath:schema.ldif
-spring.ldap.embedded.port=12345
```
andymarke: Just find any free port so tests can run while the application is up

build.gradle
```diff
@@ -125,5 +125,5 @@ test {
         showCauses = true
     }
     minHeapSize = "128m"
-    maxHeapSize = "1516m"
+    maxHeapSize = "1024m"
```
andymarke: decrease to avoid system OOM

PR #91, opened 2019-11-25
DT-245 add endpoint to current offender managers

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -102,9 +105,7 @@ public OffenderService(OffenderRepository offenderRepository, OffenderTransforme
 
         if (offenderIds == null) {
             log.error("Call to offenderRepository.listOffenderIds {}, {} returned a null list", pageSize, page);
-        }
-
-        if (offenderIds.contains(null)) {
+        } else if (offenderIds.contains(null)) {
```
andymarke: compile warning fix

src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -142,4 +143,27 @@ public Long getOffenderCount() {
...
+                                .collect(Collectors.toList()),
+                        offender.getPrisonOffenderManagers()
+                                .stream()
+                                .filter(uk.gov.justice.digital.delius.jpa.standard.entity.PrisonOffenderManager::isActive)
+                                .map(offenderManagerTransformer::offenderManagerOf)
+                                .collect(Collectors.toList())
+                ) );
+    }
+
+     private static <T> List<T> combine(List<T> first, List<T> second) {
```
andymarke: couldn't find anything in JDK that merged two lists in a functional way

petergphillips: agreed, that's the way I normally do it

PR #93, opened 2019-11-28
DT-337 recall by crn

mikehalmamoj:
Add the crn version of the recall/release endpoint, and check for offender not found.
Lacking any consumers and returning no data, this can be safely merged into master once approved.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -173,7 +173,7 @@ public class OffendersResource {
     }
 
     @ApiOperation(
-            value = "Returns the latest recall and release details for an offender",
+            value = "WARNING: This is a work in progress!  Returns the latest recall and release details for an offender", // TODO DT-337 remove warning when finished
             notes = "Accepts a NOMIS offender nomsNumber in the format A9999AA")
     @ApiResponses(
             value = {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -172,7 +173,7 @@
     }
 
     @ApiOperation(
-            value = "Returns the latest recall and release details for an offender",
+            value = "WARNING: This is a work in progress!  Returns the latest recall and release details for an offender", // TODO DT-337 remove warning when finished
```
brightonsbox: Could maybe use @ApiIgnore("WIP") so that it doesn't show in swagger documentation until it's ready?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -188,7 +189,31 @@
...
+                    @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
+                    @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+            })
+    @GetMapping(path = "/offenders/crn/{crn}/release")
+    public ResponseEntity<OffenderLatestRecall> getLatestRecallAndReleaseForOffenderByCrn(
+            @ApiParam(name = "crn", value = "CRN for the offender", example = "X320741", required = true)
+            @NotNull
+            @PathVariable(value = "crn") final String crn) {
+
+        return offenderService.offenderIdOfCrn(crn)
```
andymarke: So you can now simplify this to returning a plain `OffenderLatestRecall ` rather than a `ResponseEntity ` and throw a `NotFoundException` which will be processed our exception handler to return a 404 automatically

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
mikehalmamoj: Good point, but I think we're intending to add the API to the swagger doc as a heads up to other teams.

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
mikehalmamoj: Cool I'll change to this method in the next PR.

PR #94, opened 2019-11-29
DT-367 fix user details not found

petergphillips:


src/main/java/uk/gov/justice/digital/delius/ldap/repository/LdapRepository.java
```diff
@@ -62,74 +58,73 @@ public LdapRepository(LdapTemplate ldapTemplate) {
...
         return ldapTemplate.search(query().where(field).is(value), contextMapper);
     }
 
-    public boolean authenticateUser(String user, String password) {
+    public boolean authenticateUser(final String user, final String password) {
         return ldapTemplate.authenticate(ldapUserBase, "(uid=" + user + ")", password);
     }
 
-    public Optional<NDeliusUser> getDeliusUser(String username) {
-        val nDeliusUser = ldapTemplate.findOne(byUsername(username), NDeliusUser.class);
```
petergphillips: `findOne` throws an Entity not found exception, which then bubbles up

PR #95, opened 2019-11-29
DT-337 Offender release/recall - get custodial event

mikehalmamoj:
The point of this PR is to start implementing OffenderService#getOffenderLatestRecall and making ConvictionService#getActiveCustodialEvent public.

However there is also a refactor of SingleActiveCustodyConvictionNotFoundException which has been moved into SecureControllerAdvice so there's a lot of noise from that change.


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -211,9 +208,13 @@
             @NotNull
             @PathVariable(value = "crn") final String crn) {
 
-        return offenderService.offenderIdOfCrn(crn)
+        return getOffenderLatestRecall(offenderService.offenderIdOfCrn(crn));
+    }
+
+    private ResponseEntity<OffenderLatestRecall> getOffenderLatestRecall(Optional<Long> maybeOffenderId) {
```
andymarke: Could convert this return `OffenderLatestRecall ` rather than `ResponseEntity<OffenderLatestRecall>`

PR #96, opened 2019-11-29
DT-337 Offender release/recall - custody service stub

mikehalmamoj:
Creates a stub for a new custody service which finds the Custody record from a custodial event and plumbs the stub into the release/recall API.
UPDATE: Following code review it was determined that a new Custody Service was not needed and the code was added to the offender service.


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -188,7 +188,7 @@
             @NotNull
             @PathVariable(value = "nomsNumber") final String nomsNumber) {
 
-        return getOffenderLatestRecall(offenderService.offenderIdOfNomsNumber(nomsNumber));
+        return new ResponseEntity<>(getOffenderLatestRecall(offenderService.offenderIdOfNomsNumber(nomsNumber)), HttpStatus.OK);
```
andymarke: Can we change the return type to drop the ResponseEntity?

src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -169,6 +170,7 @@ public OffenderLatestRecall getOffenderLatestRecall(Long offenderId) {
         Offender offender = offenderRepository.findByOffenderId(offenderId)
                 .orElseThrow(() -> new NotFoundException("Offender not found"));
         uk.gov.justice.digital.delius.jpa.standard.entity.Event activeCustodialEvent = convictionService.getActiveCustodialEvent(offender.getOffenderId());
+        uk.gov.justice.digital.delius.jpa.standard.entity.Custody custody = custodyService.findCustodyFromCustodialEvent(activeCustodialEvent);
```
andymarke: Not sure we need a service to this, isn't just `Optional.of(activeCustodialEvent.getDisposal()).flatMap(d -> d.getCustody()).orElseThrow()`  

in other words the model contains it without doing further lookups?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
mikehalmamoj: Ah I see - we don't need the ResponseEntity as we are only handling the happy path here.  Yep, will do.

src/test/java/uk/gov/justice/digital/delius/OffendersResource_GetLatestRecallAndReleaseForOffender.java
```diff
@@ -234,6 +242,44 @@ public void getLatestRecallAndReleaseForOffender_notSingleCustodyEvent_returnsDe
...
+                .contentType(APPLICATION_JSON_VALUE)
+                .when()
+                .get("/offenders/crn/X320741/release")
+                .then()
+                .statusCode(400);
+    }
+
+    @Test
+    public void getLatestRecallAndReleaseForOffender_custodyRecordNotFound_returnsEventIdInErrorMessage() {
+        org.mockito.BDDMockito.given(mockOffenderService.offenderIdOfCrn(anyString()))
```
petergphillips: any reason for fully qualifying `BDDMockito` here?

mikehalmamoj: Just because we're using RestAssured's given to test the GET call.  Unless there's a better way I'm missing?

PR #98, opened 2019-12-03
DT-337 Release/recall model

mikehalmamoj:
Add the release and recall entities to the JPA model.
Adjust the API model to remove details that are not available in the JPA model.
Add transformers for the release and recall entities.


src/main/java/uk/gov/justice/digital/delius/data/api/OffenderRelease.java
```diff
@@ -15,8 +15,6 @@
 public class OffenderRelease {
     @ApiModelProperty(value = "The date the release occurred", example = "2019-11-26")
     private LocalDate date;
-    @ApiModelProperty(value = "The reason for the release")
```
andymarke: Would ReleaseType not be a the release reason? Maybe look at the values or change the model to be `releaseType` rather than `reason`

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Custody.java
```diff
@@ -40,4 +42,11 @@
     @OneToMany(cascade = {CascadeType.ALL}, orphanRemoval=true, mappedBy = "custody")
     private List<KeyDate> keyDates;
 
+    @OneToMany
+    @JoinColumn(name = "CUSTODY_ID")
+    private List<Release> releases;
+
+    public Optional<Release> findLatestRelease() {
+        return this.getReleases().stream().max(Comparator.comparing(Release::getActualReleaseDate));
```
andymarke: Maybe filter by `softDeleted` though not sure if if this is ever set by Delius

mikehalmamoj: I've dealt with the soft delete flag on the Release and Recall entities.  However I've left the custodial event / disposal / custody todo in the next piece of work as I will be refactoring some other stuff in that area.

PR #102, opened 2019-12-04
DT-337 offender recall service

mikehalmamoj:
Completes OffenderService#getOffenderLatestRecall, so the new API is now available.

I split the OffenderResource tests into 2...  the API test is a unit test that stands up the controller and controller advice, allowing us to use mocked data to comprehensively test the APIs contract.  The integration test stands up the entire application allowing us to test a few cases in full.

NOTE: This is the final PR for DT-337 and the endpoints now return data so this needs a test.


src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -137,17 +142,21 @@ public Long getOffenderCount() {
...
+        Event actualCustodialEvent = convictionService.getActiveCustodialEvent(offenderId);
+        Custody custody = findCustodyOrThrow(actualCustodialEvent);
+        return custody.findLatestRelease()
+                .map(releaseTransformer::offenderLatestRecallOf)
+                .orElse(OffenderLatestRecall.NO_RELEASE);
+    }
+
+    private Custody findCustodyOrThrow(Event activeCustodialEvent) {
+        return Optional.ofNullable(activeCustodialEvent.getDisposal())
+                .filter(not(Disposal::isSoftDeleted))
```
andymarke: I think we should also add this filtering to `convictionService.getActiveCustodialEvent` if not there already.

mikehalmamoj: This already occurs in ConvictionService#activeCustodyEvents.

PR #105, opened 2019-12-06
DT-400 split allocate pom tests

mikehalmamoj:
Split out the allocate POM API validation tests so we can turn them into unit tests


src/main/java/uk/gov/justice/digital/delius/data/api/ProbationArea.java
```diff
@@ -17,7 +17,7 @@
     private Long probationAreaId;
     @ApiModelProperty(value = "area code", example = "N01")
     private String code;
-    @ApiModelProperty(value = "area code", example = "N01")
+    @ApiModelProperty(value = "description", example = "NPS North West")
```
andymarke: Thanks!

PR #106, opened 2019-12-06
DT-401 minor swagger improvements

andymarke:


src/main/java/uk/gov/justice/digital/delius/data/api/ProbationArea.java
```diff
@@ -15,8 +15,11 @@
 public class ProbationArea {
     @ApiModelProperty(required = true)
     private Long probationAreaId;
+    @ApiModelProperty(value = "area code", example = "N01")
     private String code;
+    @ApiModelProperty(value = "area code", example = "N01")
```
andrewrlee: should this be description?

andymarke: Thanks I'll fix in next PR

PR #114, opened 2019-12-09
DT-403 Create Contact when Responsible Officer changes

andymarke:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ResponsibleOfficer.java
```diff
@@ -48,10 +48,10 @@
 
     @Column(name = "START_DATE")
     @Builder.Default
-    private LocalDate startDate = LocalDate.now();
```
andymarke: Delius reports the time that a RO is assigned hence the change

PR #117, opened 2019-12-31
DCS-297 Adding 'secure' endpoints to allow retrieving lists of probat…

andrewrlee:
…ion areas and LDUs


src/main/java/uk/gov/justice/digital/delius/service/ReferenceDataService.java
```diff
@@ -62,4 +68,19 @@ private StandardReference pomAllocationTransferReason(String reason) {
         return standardReferenceRepository.findByCodeAndCodeSetName(reason, POM_ALLOCATION_REASON_DATASET)
                 .orElseThrow(() -> new RuntimeException(String.format("No pom allocation reason found for %s", reason)));
     }
+
+    public Page<KeyValue> getProbationAreasCodes(boolean restrictActive) {
```
andrewrlee: maybe a bit weird and happy to change - returning Page even though not currently supporting pagination as its friendlier for backwards compatibility reasons to return an object rather than a list, and could potentially add pagination support at a later date?

src/main/java/uk/gov/justice/digital/delius/controller/secure/ReferenceDataResource.java
```diff
@@ -0,0 +1,50 @@
...
+@RequestMapping(value = "secure", produces = MediaType.APPLICATION_JSON_VALUE)
+@AllArgsConstructor
+@RestController
+@PreAuthorize("hasRole('ROLE_COMMUNITY')")
+public class ReferenceDataResource {
+
+    private final ReferenceDataService referenceDataService;
+
+    @GetMapping("/probationAreas")
+    public Page<KeyValue> getProbationAreaCodes(final @RequestParam(name = "active", required = false) boolean restrictActive) {
```
andrewrlee: This is only returning the code and description at the moment rather than the detailed view of each area and all of it's teams

src/main/java/uk/gov/justice/digital/delius/service/ReferenceDataService.java
```diff
@@ -62,4 +68,19 @@ private StandardReference pomAllocationTransferReason(String reason) {
...
+        final var filter = ProbationAreaFilter.builder().restrictActive(restrictActive).build();
+
+        return probationAreaRepository.findAll(filter).stream()
+                .map(area -> new KeyValue(area.getCode(), area.getDescription()))
+                .collect(collectingAndThen(toList(), PageImpl::new));
+    }
+
+    public Page<KeyValue> getLocalDeliveryUnitsForProbationArea(String code) {
+        return probationAreaRepository.findByCode(code).stream()
```
andymarke: Slightly confused why this returns a List I would have thought it would be an Optional
`probationAreaRepository.findByCode(code)`

src/test/java/uk/gov/justice/digital/delius/ReferenceDataResourceTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+import java.util.List;
+
+import static io.restassured.RestAssured.given;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@RunWith(SpringJUnit4ClassRunner.class)
+@ActiveProfiles("dev-seed")
+public class ReferenceDataResourceTest {
```
andymarke: Interestingly we have a "API" test for the non-secure version. Maybe after this PR we should think about restructuring tests so we can differentiate between  secure/non-secure mock/integration tests.

src/main/java/uk/gov/justice/digital/delius/service/ReferenceDataService.java
andrewrlee: [stream](https://docs.oracle.com/javase/9/docs/api/java/util/Optional.html#stream--) is a new method on Optional

PR #118, opened 2020-01-06
DT-449 speed up auth

mikehalmamoj:
Fix a performance bug where logging in takes several seconds due to an excessive search scope.


src/main/java/uk/gov/justice/digital/delius/ldap/repository/LdapRepository.java
```diff
@@ -56,16 +63,12 @@ public LdapRepository(final LdapTemplate ldapTemplate) {
         };
     }
 
-    public List<Map<String, String>> searchByFieldAndValue(final String field, final String value) {
```
mikehalmamoj: This method was only being used by UserController which was itself was only being used to diagnose issues (it has a strange Spring profile and it never appears in AppInsights in production).  Hence I deleted this method and UserController.

src/main/java/uk/gov/justice/digital/delius/ldap/repository/LdapRepository.java
```diff
@@ -56,16 +63,12 @@ public LdapRepository(final LdapTemplate ldapTemplate) {
...
-        return ldapTemplate.search(query().where(field).is(value), contextMapper);
-    }
-
```
mikehalmamoj: This method was only being used by UserController which was itself was only being used to diagnose issues (it has a strange Spring profile and it never appears in AppInsights in production). Hence I deleted this method and UserController.

PR #121, opened 2020-01-15
DCS-309 Adding endpoint to allow setting roles

andrewrlee:
Adding new endpoint to allow idempotent assigning of roles.

Hit a limitation of the embedded ldap server which doesn't support aliases which is how delius model roles.
So locally we cannot retrieve descriptions of roles - we don't think descriptions are used by any clients so removing.

Also moving from using uid as part of a user's dn to use cn instead. This is consistent with the delius codebase and we
think it should lead to equivalent behaviour.

Also moving OffenderResource to have a consistent swagger description as the other secure endpoints - as it's currently hard to see in the docs.


src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
```diff
@@ -59,6 +59,24 @@ public UserDetails findUser(@ApiParam(name = "username", value = "LDAP username"
                 .orElseThrow(() -> new NotFoundException(String.format("User with username %s", username)));
     }
 
+    @ApiOperation(
+            value = "Find user details of a user held in Delius Identity (LDAP)",
```
andymarke: minor, but worth amending the description

src/main/java/uk/gov/justice/digital/delius/ldap/repository/LdapRepository.java
```diff
@@ -81,13 +66,42 @@ public boolean authenticateUser(final String user, final String password) {
...
                                             .builder()
                                             .cn(attributes.get("cn").get().toString())
-                                            .description(attributes.get("description").get().toString())
                                             .build());
 
             return user.toBuilder().roles(roles).build();
         });
     }
 
+    public void addRole(String username, String roleId)  {
```
andymarke: Do you think we need a check for if the user already has this role assigned? LDAP might allow both duplicate roles to exist that could cause issues with removal

src/test/java/uk/gov/justice/digital/delius/ldap/repository/LdapTestConfig.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.digital.delius.ldap.repository;
```
andymarke: Do we need this class?

andrewrlee: Added this in error, removed now

src/main/java/uk/gov/justice/digital/delius/ldap/repository/LdapRepository.java
andrewrlee: as discussed - it does a rebind on a DN so should be idempotent.

src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
andrewrlee: fixed

PR #124, opened 2020-01-20
DCS-306 Adding endpoint to allow retrieval of teams for a given LDU

andrewrlee:


src/main/java/uk/gov/justice/digital/delius/service/ReferenceDataService.java
```diff
@@ -84,4 +85,18 @@ private StandardReference pomAllocationTransferReason(String reason) {
                 .map(unit -> new KeyValue(unit.getCode(), unit.getDescription()))
                 .collect(collectingAndThen(toList(), PageImpl::new));
     }
+
+    public Page<KeyValue> getTeamsForLocalDeliveryUnit(String code, String lduCode) {
```
mikehalmamoj: Is this missing a unit test?

andrewrlee: Added missing service tests

PR #126, opened 2020-01-22
DCS-329 Fixing returned data from new reference data endpoints

andrewrlee:
So now:
 * Only including selectable LDUs from new endpoints
 * Also now returning all teams for a given LDU code rather than just those for the first LDU with that specific code


src/test/java/uk/gov/justice/digital/delius/service/ReferenceDataServiceTest.java
```diff
@@ -147,9 +148,65 @@ public void getTeamsForLocalDeliveryUnit_LduDoesntExist() {
                 .hasMessage("Could not find local delivery unit in probation area: 'NO2', with code: 'LDU-2'");
     }
 
+    @Test
+    public void getTeamsForLocalDeliveryUnit_unselectableLdusAreNotReturned() {
+        when(probationAreaRepository.findByCode(aProbationArea().getCode())).thenReturn(
+                Optional.of(
```
petergphillips: this would look a lot nicer in :kotlin: without the builders :grinning:

PR #134, opened 2020-01-29
DT-463 Add custody history record when location change

andymarke:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Custody.java
```diff
@@ -72,10 +72,10 @@ public String getCode() {
     private List<Release> releases;
 
     @Column(name = "STATUS_CHANGE_DATE")
-    private LocalDateTime statusChangeDate;
+    private LocalDate statusChangeDate;
```
andymarke: Delius currently only sets the date part of these columns so lets do the same

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -8,44 +8,55 @@
...
+        this.custodyHistoryRepository = custodyHistoryRepository;
+        this.referenceDataService = referenceDataService;
     }
 
     @Transactional
     public Custody updateCustody(final String nomsNumber,
                                  final String bookingNumber,
                                  final UpdateCustody updateCustody) {
-        final var maybeOffender = offenderService.getOffenderByNomsNumber(nomsNumber);
+        final var maybeOffender = offenderRepository.findByNomsNumber(nomsNumber);
```
andymarke: switched to use raw entity since we need it later for the custody history entity

src/test/java/uk/gov/justice/digital/delius/integration/secure/CustodyUpdateAPITest.java
```diff
@@ -99,11 +102,22 @@ public void updatePrisonInstitution() throws JsonProcessingException {
         assertThat(custody.getInstitution().getNomsPrisonInstitutionCode()).isEqualTo("MDI");
         verify(telemetryClient).trackEvent(eq("P2PTransferPrisonUpdated"), any(), isNull());
 
+        //custody record should have been updated
         final var custodyRecord = jdbcTemplate.query(
                 "SELECT * from CUSTODY where PRISONER_NUMBER = ?",
                 List.of(PRISON_BOOKING_NUMBER).toArray(),
                 new ColumnMapRowMapper()).get(0);
         assertThat(toLocalDateTime(custodyRecord.get("LAST_UPDATED_DATETIME"))).isCloseTo(LocalDateTime.now(), within(10, ChronoUnit.SECONDS));
+        assertThat(toLocalDateTime(custodyRecord.get("LOCATION_CHANGE_DATE")).toLocalDate()).isEqualTo(LocalDate.now());
```
andymarke: we have no end points for these entities so use raw SQL to test our side effects are correct

mikehalmamoj: Which is fine, but if you wanted an alternative you could use a SpyBean for the repository.

PR #136, opened 2020-01-29
DT-463 Notify SPG of custody change and status change

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/SpgNotificationService.java
```diff
@@ -34,11 +34,13 @@ public SpgNotificationService(BusinessInteractionRepository businessInteractionR
 
     enum NotificationEvents {
         INSERT_EVENT("CWBI006"),
+        AMEND_EVENT("SEBI014"),
```
andymarke: Codes taken directly from Delius

src/test/java/uk/gov/justice/digital/delius/service/CustodyServiceTest.java
```diff
@@ -161,6 +162,15 @@ public void willCreateCustodyHistoryChangeLocationEvent() {
         assertThat(custodyHistoryEvent.getWhen()).isEqualTo(LocalDate.now());
     }
 
+    @Test
+    public void willNotifySPGOfCustodyChange() {
+        when(institutionRepository.findByNomisCdeCode("MDI")).thenReturn(Optional.of(EntityHelper.anInstitution().toBuilder().description("HMP Highland").build()));
+
+        custodyService.updateCustody("G9542VP", "44463B", UpdateCustody.builder().nomsPrisonInstitutionCode("MDI").build());
+
+        verify(spgNotificationService).notifyUpdateOfCustody(any(), any());
```
brightonsbox: Could maybe check the expected event and offender parameters are used here?

andymarke: I did think that but it is an odd thing since they are not passed into service, and are the result of other service calls, so it ends up testing the results of mocks are fed into this method, the whole thing than looks odd and not very revealing.

PR #141, opened 2020-02-04
DT-463 SPG should always be called twice (and removed stupid logging)

andymarke:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/SpgNotification.java
```diff
@@ -58,7 +58,8 @@
     @Column(name = "SPG_INTERCHANGE_STATUS_ID")
     private Long spgInterchangeStatusId;
     @Column(name = "ROW_VERSION")
-    private Long rowVersion;
+    @Builder.Default
+    private Long rowVersion = 1L;
```
andymarke: previously default to 0, but Delius starts with 1

PR #142, opened 2020-02-06
pat-518 new endpoint to retrieve convictions by crn

lbennett-moj:


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -290,5 +290,25 @@ public OffenderDetail getOffenderDetailByNomsNumber(final @PathVariable("nomsNum
...
+            value = {
+                    @ApiResponse(code = 200, message = "OK", response = Conviction.class, responseContainer = "List"),
+                    @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+                    @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
+                    @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
+                    @ApiResponse(code = 404, message = "The offender is not found", response = ErrorResponse.class),
+                    @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+            })
+    @GetMapping(path = "/offenders/crn/{crn}/convictions")
+    public ResponseEntity<List<Conviction>> getConvictionsForOffenderByCrn(
```
petergphillips: would be better to just return the list of convictions and let the global exception mapper handle the not found

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderConvictionsByCrn.java
```diff
@@ -0,0 +1,77 @@
...
+                .contentType(APPLICATION_JSON_VALUE)
+                .when()
+                .get("/offenders/crn/X320741/convictions")
+                .then()
+                .statusCode(200)
+                .extract()
+                .body()
+                .as(Conviction[].class);
+
+        final var conviction = Stream.of(convictions).filter(Conviction::getActive).findAny().orElseThrow();
```
petergphillips: would be better to just compare the json rather than rehydrating into the same class.  That way if the api changes we are making a conscious decision to change it

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
lbennett-moj: sounds good - I'll change it

petergphillips: so the return type should be `List<Conviction>`

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -290,5 +290,25 @@ public OffenderDetail getOffenderDetailByNomsNumber(final @PathVariable("nomsNum
...
+            @ApiParam(name = "crn", value = "CRN for the offender", example = "A123456", required = true)
+            @NotNull @PathVariable(value = "crn") final String crn) {
+
+        return offenderService.offenderIdOfCrn(crn)
+                .map(offenderId -> new ResponseEntity<>(convictionService.convictionsFor(offenderId), HttpStatus.OK))
```
petergphillips: and this can just be
```java
.map(offenderId -> convictionService.convictionsFor(offenderId))
```

PR #143, opened 2020-02-07
DT-547 When creating a CONTACT use ACTIVE order manger

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ContactService.java
```diff
@@ -76,8 +76,10 @@ public void addContactForResponsibleOfficerChange(PrisonOffenderManager newPriso
     @Transactional
     public void addContactForPrisonLocationChange(Offender offender, Event event) {
         final var contactType = contactTypeForPrisonLocationChange();
-        // same as Delius get first Order manager
-        final var mayBeOrderManager = event.getOrderManagers().stream().findFirst();
```
andymarke: No idea how this works in Delius (which is where I copied the code) since it will get old inactive managers; any way here is the simple fix

PR #144, opened 2020-02-07
DT-549 Add additional filter to exclude prisons from probation list

andymarke:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ProbationArea.java
```diff
@@ -54,4 +54,7 @@
     @Column(name = "PRIVATE")
     private Long privateSector;
 
+    @Column(name = "ESTABLISHMENT", updatable = false, insertable = false)
```
andymarke: mapped above in a composite key hence the `updatable=false`

PR #146, opened 2020-02-14
DT-466 Add do nothing end point for updating booking number

andymarke:
End point does *nothing* but needed a placeholder

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
@@ -8,6 +8,7 @@ import org.springframework.transaction.annotation.Transactional;
 import uk.gov.justice.digital.delius.controller.NotFoundException;
 import uk.gov.justice.digital.delius.data.api.Custody;
 import uk.gov.justice.digital.delius.data.api.UpdateCustody;
+import uk.gov.justice.digital.delius.data.api.UpdateCustodyBookingNumber;
 import uk.gov.justice.digital.delius.jpa.standard.entity.CustodyHistory;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -107,6 +108,16 @@ public Custody updateCustody(final String nomsNumber,
         });
     }
 
+
+    public Custody updateCustodyBookingNumber(String nomsNumber, UpdateCustodyBookingNumber updateCustodyBookingNumber) {
+        log.info("Request to update booking number for {}", nomsNumber);
+        return Custody
+                .builder()
+                .bookingNumber(updateCustodyBookingNumber.getBookingNumber())
```
andymarke: look; does nothing :-)

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -126,7 +137,7 @@ private Event updateInstitutionOnEvent(Offender offender, Event event, RInstitut
             if (custody.isAboutToEnterCustody()) {
                 custody.setStatusChangeDate(LocalDate.now());
                 custody.setCustodialStatus(referenceDataService.getInCustodyCustodyStatus());
-                saveCustodyStatusChangeCustodyHistoryEvent(offender, custody, institution);
```
andymarke: unrelated, removing compile warnings

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
Mjwillis: Lol

PR #147, opened 2020-02-17
PAT-554 Add 'selectable' column to the ProbationArea entity and use it to determine if a Probation Area is active

yasinmustafa:


src/test/java/uk/gov/justice/digital/delius/integration/secure/ReferenceDataResourceTest.java
```diff
@@ -106,7 +106,7 @@ public void canGetOnlyActiveProbationAreasExcludingPrisons() {
                 .jsonPath().getList("content", KeyValue.class);
 
         assertThat(probationAreas)
-                .hasSize(73)
+                .hasSize(45)
```
andymarke: Numbers looking much better even though this is test data!

PR #148, opened 2020-02-17
DT-466 Lookup offender and conviction

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -132,6 +146,28 @@ public Conviction addCourtCaseFor(Long offenderId, CourtCase courtCase) {
...
+            case 1:
+                return events.stream().findFirst();
+            default:
+                throw new DuplicateConvictionsForSentenceDateException(events.size());
+        }
+    }
+
+    private boolean didSentenceStartAroundDate(Event event, LocalDate sentenceStartDate) {
+        // typically used to match start dates in NOMIS and Delius which may be out by a few days
+        return DAYS.between(event.getDisposal().getStartDate(), sentenceStartDate) <= 7;
```
andymarke: Further analysis required to see if 7 days is the correct leniency between NOMIS and Delius data

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -108,13 +109,42 @@ public Custody updateCustody(final String nomsNumber,
...
+        final var maybeOffender = offenderRepository.findByNomsNumber(nomsNumber);
+        final var telemetryProperties = Map.of("offenderNo", nomsNumber,
+                "bookingNumber", updateCustodyBookingNumber.getBookingNumber(),
+                "sentenceStartDate", updateCustodyBookingNumber.getSentenceStartDate().format(DateTimeFormatter.ISO_DATE));
+
+        return maybeOffender.map(offender -> {
+            try {
+                final var maybeEvent = convictionService.getSingleActiveConvictionIdByOffenderIdAndCloseToSentenceDate(offender.getOffenderId(), updateCustodyBookingNumber.getSentenceStartDate());
+                return maybeEvent.map(event -> {
+                    // update booking Number
```
andymarke: All the TOD bits in comments, so service still does nothing

src/test/java/uk/gov/justice/digital/delius/service/ConvictionServiceTest.java
```diff
@@ -29,7 +30,7 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-@RunWith(SpringRunner.class)
+@ExtendWith(SpringExtension.class)
```
andymarke: Refactored to JUnit 5 and introduced nested hence the huge change - really just moved tests into appropriate nested tests

src/test/java/uk/gov/justice/digital/delius/service/CustodyServiceTest.java
```diff
@@ -1,12 +1,14 @@
 package uk.gov.justice.digital.delius.service;
 
 import com.microsoft.applicationinsights.TelemetryClient;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
```
andymarke: Also refactored to JUnit 5 and introduced nested hence the huge change - really just moved tests into appropriate nested tests, and added new tests for this change

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
matt-crowson-moj: Perhaps include TODO in the comment?  Just thinking if you want to search for outstanding stuff.  IDE's can then help out hear searching or even grepping for example....just a thought.

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
mikehalmamoj: Should 7 be a constant?

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -108,13 +109,43 @@ public Custody updateCustody(final String nomsNumber,
...
-                .build();
+        final var maybeOffender = offenderRepository.findByNomsNumber(nomsNumber);
+        final var telemetryProperties = Map.of("offenderNo", nomsNumber,
+                "bookingNumber", updateCustodyBookingNumber.getBookingNumber(),
+                "sentenceStartDate", updateCustodyBookingNumber.getSentenceStartDate().format(DateTimeFormatter.ISO_DATE));
+
+        return maybeOffender.map(offender -> {
+            try {
+                final var maybeEvent = convictionService.getSingleActiveConvictionIdByOffenderIdAndCloseToSentenceDate(offender.getOffenderId(), updateCustodyBookingNumber.getSentenceStartDate());
+                return maybeEvent.map(event -> {
```
andrewrlee: could do the `orElseThrow` where variable declared - then you wouldn't have to do the `map` and so could avoid large amounts of nesting?

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -51,6 +53,18 @@ public int getConvictionCount() {
             return convictionCount;
         }
     }
+    public static class DuplicateConvictionsForSentenceDateException extends Exception {
```
matt-crowson-moj: This is for my benefit really.

Lots of checked exceptions for controlling flow?  Just seems a bit old school.  Is it historic?  Sorry don't know what is new versus existing...

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
andymarke: Not sure, the problem is essentially we have a few Optionals, but for each one that we done have a value I want to do something slightly different with the Telemetry event.

`maybeA.elseThrow({do something).somethingElse().elseThrow({do something else})` but I will give it a go next revision.

Luckily all these Telemetry events will be removed once I've done and rolled out, so it should become a flat pipeline.

Thanks

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
andymarke: Good point; we have 3 states; 

1. I have found the thing
2. I have not found the thing
3. I have found loads of things when there should only be 1.

Not sure the best way to handle that other than am Optional and exception.

I suppose I could have a `Result` type class instead. Any better ideas @matt-crowson-moj

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
andymarke: Good, point, although I'm doing the TODOs as we speak

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
andymarke: Good point, I'll do that next PR

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
matt-crowson-moj: If only java had such a class (Kotlin caters this one).  Of course there are 3rd party libs that support this sort of thing providing Try and Result objects...

PR #149, opened 2020-02-18
DT-466 PR Comments, change flow (Optional and Error)

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/Result.java
```diff
@@ -0,0 +1,38 @@
...
+        this.value = value;
+    }
+    private Result(X error) {
+        this.error = error;
+    }
+
+    public T onError(Function<X, ? extends RuntimeException> exceptionSupplier) {
+        if (value != null) {
+            return value;
+        } else {
```
matt-crowson-moj: Meh you don't need the else....

src/main/java/uk/gov/justice/digital/delius/service/Result.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.justice.digital.delius.service;
+
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+
+public class Result<T, X extends Throwable> {
```
matt-crowson-moj: You could argue there should be a unit test for this class ;)

src/main/java/uk/gov/justice/digital/delius/service/Result.java
andymarke: Interestingly JDK do use else - from Optional
`    public <X extends Throwable> T orElseThrow(Supplier<? extends X> exceptionSupplier) throws X {
        if (value != null) {
            return value;
        } else {
            throw exceptionSupplier.get();
        }
    }`

I was taught not to - oh well ...

PR #150, opened 2020-02-19
DT 466 Add feature switch

andymarke:


src/test/java/uk/gov/justice/digital/delius/service/ResultTest.java
```diff
@@ -0,0 +1,59 @@
+package uk.gov.justice.digital.delius.service;
```
andymarke: Also added missing  unrelated test

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -52,6 +55,7 @@ public CustodyService(
             OffenderManagerService offenderManagerService,
             ContactService contactService) {
         this.updateCustodyFeatureSwitch = updateCustodyFeatureSwitch;
+        this.updateBooingNumberFeatureSwitch = updateBooingNumberFeatureSwitch;
```
mikehalmamoj: Great typo!

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -52,6 +55,7 @@ public CustodyService(
             OffenderManagerService offenderManagerService,
             ContactService contactService) {
```
mikehalmamoj: Possibly not one for this PR, but the number of dependencies smells like this class might need breaking up?

PR #151, opened 2020-02-20
Do the Delius Updates

andymarke:
We replicate Delius when updating booking number which includes

- Updating the booking number (aka prison number) on Custody record
- Updating the most recent number of the Offender record
- Recreating all the associated OFFENDER_PRISONER records
- Creating a CONTACT record
- Notify CRCs via the SPG of the updated EVENT

All rules have been lifted from existing Delius legacy application

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/OffenderPrisonerService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/OffenderPrisonerService.java
@@ -1,0 +1,65 @@
+package uk.gov.justice.digital.delius.service;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import uk.gov.justice.digital.delius.jpa.standard.entity.*;
+import uk.gov.justice.digital.delius.jpa.standard.repository.EventRepository;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/digital/delius/service/OffenderPrisonerService.java
```diff
@@ -0,0 +1,65 @@
...
+    public Offender refreshOffenderPrisonersFor(Offender offender) {
+        final var offenderId = offender.getOffenderId();
+        final var events = eventRepository.findByOffenderId(offenderId);
+        offenderPrisonerRepository.deleteAllByOffenderId(offenderId);
+        offenderPrisonerRepository.saveAll(createOffenderPrisonersFromEvents(offenderId, events));
+        offender.setMostRecentPrisonerNumber(getPrisonNumberFromLatestEvent(events));
+        return offender;
+    }
+
+    public String getPrisonNumberFromLatestEvent(List<Event> events) {
```
matt-crowson-moj: Just wondering why not `Optional<String>`?

andymarke: we have to set `offender.setMostRecentPrisonerNumber()` if we have a prisoner number or not, or to but it another way if there was no value we would clear it by setting it to blank hence the `orElse("")`

PR #152, opened 2020-02-20
DT-466 Do nothing when the booking number is unchanged (except Telemetry)

andymarke:


src/test/java/uk/gov/justice/digital/delius/service/CustodyServiceTest.java
```diff
@@ -482,6 +482,68 @@ void nothingWillBeUpdated() {
...
+                    final var disposal = aCustodyEvent().getDisposal().toBuilder().custody(custody).build();
+                    when(convictionService.getSingleActiveConvictionIdByOffenderIdAndCloseToSentenceDate(anyLong(), any()))
+                            .thenReturn(Result.of(Optional.of(aCustodyEvent().toBuilder().disposal(disposal).build())));
+                }
+
+                @Test
+                void nothingWillBeUpdated() {
+                    var custody = custodyService.updateCustodyBookingNumber("G9542VP", updateCustodyBookingNumber);
+
+                    verify(offenderPrisonerService, never()).refreshOffenderPrisonersFor(any());
```
brightonsbox: Not sure if you want to verify these are called in the cases where number is new or overwritten?

andymarke: They sort of are above, but I think the test isn't clear to show that. Will change in next PR

PR #156, opened 2020-03-03
PAT-594 Add endpoint to retrieve user details given a list of multiple user names

yasinmustafa:


yasinmustafa:
superseding with another PR

PR #160, opened 2020-03-05
PAT-594 Add an endpoint to retrieve multiple user details /users/list/detail

yasinmustafa:


src/main/java/uk/gov/justice/digital/delius/controller/secure/UserController.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+@RestController
+@RequestMapping(value = "secure", produces = MediaType.APPLICATION_JSON_VALUE)
+@Api(tags = "User API", authorizations = {@Authorization("ROLE_COMMUNITY")})
+@AllArgsConstructor
+@PreAuthorize("hasRole('ROLE_COMMUNITY')")
```
andymarke: I think I would be tempted to protect this with the `ROLE_AUTH_DELIUS_LDAP` role instead. There can be argument that this returns sensitive information about our uses and would be consistent with the AuthenticationController that returns similar data. Obviously the pathfinder client will need to have that role added, but that is trivial

src/main/java/uk/gov/justice/digital/delius/service/UserService.java
```diff
@@ -85,6 +83,18 @@ public AccessLimitation accessLimitationOf(final String subject, final OffenderD
                         .build()).collect(toList());
     }
 
+    public Map<String, Optional<UserDetails>> getUserDetailsMap(final Set<String> usernames) {
+        return usernames.stream().collect(Collectors.toMap(username -> username, username -> ldapRepository.getDeliusUser(username).map(user ->
+                UserDetails
+                        .builder()
+                        .roles(user.getRoles().stream().map(role -> UserRole.builder().name(role.getCn()).build()).collect(toList()))
```
andymarke: I think I would be tempted to *not* add the roles, I don't think clients of this will ever need to the roles and are very internal to Delius (Auth needs to but not anyone else)

yasinmustafa: agreed - dropped and pushed up again

src/main/java/uk/gov/justice/digital/delius/controller/secure/UserController.java
yasinmustafa: agreed and changed

PR #162, opened 2020-03-05
DT-465 Flesh out controller and new feature switch

andymarke:


src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesController.java
```diff
@@ -75,19 +76,15 @@ public Custody replaceAllCustodyKeyDateByNomsNumberAndBookingNumber(final @PathV
...
-                        .postSentenceSupervisionEndDate(replaceCustodyKeyDates.getPostSentenceSupervisionEndDate())
-                        .sentenceExpiryDate(replaceCustodyKeyDates.getSentenceExpiryDate())
-                        .build())
-                .build();
+
+        var offenderId = offenderService.offenderIdOfNomsNumber(nomsNumber).orElseThrow(() -> new NotFoundException(String.format("Offender with NOMS number %s not found", nomsNumber)));
+        try {
+            var convictionId = convictionService.getSingleActiveConvictionIdByOffenderIdAndPrisonBookingNumber(offenderId, bookingNumber).orElseThrow(() -> new NotFoundException(String.format("Conviction with bookingNumber %s not found for offender with NOMS number %s", bookingNumber, nomsNumber)));
+            return convictionService.addOrReplaceOrDeleteCustodyKeyDates(offenderId, convictionId, replaceCustodyKeyDates);
+        } catch (DuplicateConvictionsForBookingNumberException e) {
```
andymarke: No 100% happy with this pattern of handling exceptions here but it is consistent with the rest of this controller (and of course does work)

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -133,7 +137,7 @@ public Conviction addCourtCaseFor(Long offenderId, CourtCase courtCase) {
         }
     }
 
-    public Optional<Event> getSingleActiveConvictionIdByOffenderIdAndPrisonBookingNumber(Long offenderId, String prisonBookingNumber) throws DuplicateConvictionsForBookingNumberException {
+    public Optional<Event> getSingleActiveConvictionByOffenderIdAndPrisonBookingNumber(Long offenderId, String prisonBookingNumber) throws DuplicateConvictionsForBookingNumberException {
```
andymarke: Rename and then add the variation below that *does* return the ID

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -222,6 +231,21 @@ public Event getActiveCustodialEvent(Long offenderId) {
...
     }
 
+    @SuppressWarnings("unused")
+    @Transactional
+    public Custody addOrReplaceOrDeleteCustodyKeyDates(Long offenderId, Long convictionId, ReplaceCustodyKeyDates replaceCustodyKeyDates) {
+        var event = eventRepository.findById(convictionId).orElseThrow();
+
+        //noinspection StatementWithEmptyBody
+        if (updateCustodyKeyDatesFeatureSwitch) {
+            // TODO the actual update
```
andymarke: The meat of this will go here - so still does nothing

src/test/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesControllerTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private OffenderService offenderService = mock(OffenderService.class);
+    private ConvictionService convictionService = mock(ConvictionService.class);
+    private ObjectMapper objectMapper = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+            .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+            .setSerializationInclusion(JsonInclude.Include.NON_ABSENT)
+            .setSerializationInclusion(JsonInclude.Include.NON_EMPTY)
+            .registerModules(new Jdk8Module(), new JavaTimeModule());
```
mikehalmamoj: Is this the right place to configure the object mapper?  It might get out of sync with the real object mapper.  (In the past I have configured object mappers in a static method just so I can reuse in unit tests)

src/test/java/uk/gov/justice/digital/delius/service/ConvictionServiceTest.java
```diff
@@ -250,11 +250,33 @@ public void convictionReturnedWhenSingleActiveConvictionMatchedAmoungstDuplicate
...
+                    aEvent()
+                            .toBuilder()
+                            .eventId(999L)
+                            .activeFlag(1L)
+                            .build()
+            ));
+
+            Optional<Long> maybeConviction = convictionService.getSingleActiveConvictionIdByOffenderIdAndPrisonBookingNumber(99L, "A12345");
+
+            assertThat(maybeConviction.orElseThrow())
```
mikehalmamoj: Just wondering what the `orElseThrow` brings to the table?

src/test/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesControllerTest.java
andymarke: Totally agree, I'll do something next time around. The problem being is we don't use Spring to inject this so we need one that matches what we do in Spring

src/test/java/uk/gov/justice/digital/delius/service/ConvictionServiceTest.java
andymarke: I suppose it is the same as writing `assertThat(maybeConviction).get().isEqualTo(999L)`  notice if I did `assertThat(maybeConviction.get()).isEqualTo(999L)` IntelliJ would warn me. So I suppose it is saying "I really expect to have a value here, but if don't throw an exception and fail"

PR #163, opened 2020-03-09
DT-465 Update and delete the actual dates

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -283,6 +301,13 @@ private String calculateNextEventNumber(Long offenderId) {
         return keyDate -> keyDate.getKeyDateType().getCodeValue().equals(typeCode);
     }
 
+    private void addOrReplaceCustodyKeyDate(Event event, String typeCode, LocalDate date)  {
+        try {
+            addOrReplaceCustodyKeyDate(event, typeCode, CreateCustodyKeyDate.builder().date(date).build());
+        } catch (CustodyTypeCodeIsNotValidException e) {
+            throw new RuntimeException(e);
```
mikehalmamoj: Is it better to have our own RuntimeException here?

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -231,21 +233,37 @@ public Event getActiveCustodialEvent(Long offenderId) {
...
+                    .getDisposal()
+                    .getCustody()
+                    .getKeyDates()
+                    .stream()
+                    .map(KeyDate::getKeyDateType)
+                    .map(StandardReference::getCodeValue)
+                    .filter(custodyManagedKeyDates::contains)  // all key dates managed by this service
+                    .filter(missingKeyDateTypesCodes::contains) // all ones missing from request
+                    .collect(toList()) // collect into new list so we can start deleting
+                    .forEach(keyDate -> deleteCustodyKeyDate(event, keyDate));
```
mikehalmamoj: Possibly move this into a method to document it better?

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
andymarke: In this case it is "really shouldn't happen" since the code is not passed from a client unlike the other API so no need to do anything more than thrown "something"

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
andymarke: I might do in next iteration; I know I have to refactor this when I deal with creating a Contact note

PR #165, opened 2020-03-10
PAT-645 Change endpoint for /users/list/detail to return a list rather than a map and associated  swagger doc changes

yasinmustafa:


src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
```diff
@@ -62,18 +63,19 @@ public void retrieveUserDetailsForMultipleUsers() throws JsonProcessingException
...
+                .as(UserDetails[].class);
 
-        Map<String, UserDetails> userDetailsMap = objectMapper.convertValue(userDetails, new TypeReference<>() {});
-        UserDetails jimSnowUserDetails = userDetailsMap.get("JimSnowLdap");
-        UserDetails sheilaHancockUserDetails = userDetailsMap.get("sheilahancocknps");
+        UserDetails jimSnowUserDetails = Stream.of(userDetails).filter(u -> u.getUsername().equals("JimSnowLdap")).findFirst().orElseThrow();
+        UserDetails sheilaHancockUserDetails = Stream.of(userDetails).filter(u -> u.getUsername().equals("sheilahancocknps")).findFirst().orElseThrow();
 
-        assertThat(userDetails.size()).isEqualTo(2);
+        assertThat(userDetails.length).isEqualTo(2);
```
petergphillips: should use `hasSize` instead

src/main/java/uk/gov/justice/digital/delius/controller/secure/UserController.java
```diff
@@ -28,18 +27,17 @@
...
             @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
             @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
             @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
             @ApiResponse(code = 404, message = "Not found", response = ErrorResponse.class),
             @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)})
     @PostMapping(path="/users/list/detail", consumes = "application/json")
-    public Map<String, Optional<UserDetails>> getUserDetailsMap(final @RequestBody Set<String> usernames){
-        log.info("getUserDetailsMap called with {}", usernames);
-        return userService.getUserDetailsMap(usernames);
+    public List<Optional<UserDetails>> getUserDetailsList(final @RequestBody Set<String> usernames){
```
petergphillips: don't think it makes much sense for this to be `Optional` and should have a wrapper object too - see https://mojdt.slack.com/archives/GCBGDGCLB/p1565767282002300?thread_ts=1565679964.000300&cid=GCBGDGCLB

src/main/java/uk/gov/justice/digital/delius/data/api/UserDetails.java
```diff
@@ -29,4 +29,6 @@
     private boolean enabled;
     @ApiModelProperty(value = "Roles For this User", position = 6, allowEmptyValue = true)
     private List<UserRole> roles;
+    @ApiModelProperty(value = "The username of the user", example = "test.user", allowEmptyValue = true, position = 7)
```
petergphillips: why are you allowing empty values - `allowEmptyValue` is set to true?

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
```diff
@@ -62,18 +63,19 @@ public void retrieveUserDetailsForMultipleUsers() throws JsonProcessingException
...
 
+        assertThat(jimSnowUserDetails).isNotNull();
```
petergphillips: why will this ever be null if you are doing an `orElseThrow` above?

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
```diff
@@ -62,18 +63,19 @@ public void retrieveUserDetailsForMultipleUsers() throws JsonProcessingException
                 .statusCode(200)
                 .extract()
                 .body()
-                .as(JsonNode.class);
+                .as(UserDetails[].class);
```
petergphillips: would be a much better test if we actually converted to a `String.class` and then compared the json instead.  That way you could compare all the data rather than just a subset of the fields.

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
yasinmustafa: agreed - done

src/main/java/uk/gov/justice/digital/delius/controller/secure/UserController.java
yasinmustafa: thanks and agreed, removed the Optional and wrapped the list in a object

src/main/java/uk/gov/justice/digital/delius/data/api/UserDetails.java
yasinmustafa: This object is used by another api call as well that doesn't set the username. Discussed this originally with Andy Marke, but will have a second chat about it. Maybe we can extend to another object.

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
yasinmustafa: agreed - done

src/main/java/uk/gov/justice/digital/delius/data/api/UserDetails.java
petergphillips: that makes sense then :+1:

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
yasinmustafa: Good point, agreed and done

src/test/java/uk/gov/justice/digital/delius/integration/secure/UserAPITest.java
```diff
@@ -62,21 +60,26 @@ public void retrieveUserDetailsForMultipleUsers() throws JsonProcessingException
...
+                       "},{" +
+                           "\"firstName\":\"Sheila\"," +
+                           "\"surname\":\"Hancock\"," +
+                           "\"email\":\"sheila.hancock@justice.gov.uk\"," +
+                           "\"enabled\":true," +
+                           "\"username\":\"sheilahancocknps\"" +
+                       "}]" +
+                       "}";
+
+        assertThat(userDetails).isEqualTo(expectedJson);
```
petergphillips: I was about to write a comment about how you can't guarantee the format of the json, but then realised that you're converting into a `JsonNode` and then converting that to a string so the format will be guaranteed :grinning: .  One alternative to this is to use jsonassert or another json library to compare the two strings...

PR #171, opened 2020-03-23
Pic 257 add unpaid work to convictions

jhevans:
Unpaid work details will now be brought back along with convictions data where present.


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Disposal.java
```diff
@@ -68,6 +68,9 @@
     @OneToMany(targetEntity = Requirement.class, mappedBy = "disposal")
     private List<Requirement> requirements;
 
+    @OneToOne(mappedBy = "disposal")
```
andymarke: Maybe check if this needs to be lazy, I think we might use Disposal a lot, not 100% sure

src/test/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformerTest.java
```diff
@@ -325,6 +328,59 @@ public void orderManagerAllocationReasonIsAlwaysNexEventCreated() {
...
+                                .build())
+                        .build())
+                .build();
+        UnpaidWork unpaidWork = transformer.convictionOf(event)
+                .getSentence()
+                .getUnpaidWork();
+
+        assertThat(unpaidWork).isNotNull();
+        assertThat(unpaidWork.getMinutesOrdered()).isEqualTo(120L);
+        assertThat(unpaidWork.getMinutesCompleted()).isEqualTo(60L);
```
chris-faulkner-mojdt: Think I'd like a test which summed the minutes across multiple UpwAppointments - if that is what getMinutesCompleted is doing

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Disposal.java
jhevans: I did have  a look at this and we can make this lazy but it won't actually bring us any benefit on it's own because this field will be retrieved automatically in all cases regardless because the transformer will always check to see if unpaid work is there. We could add an `includeUnpaidWork` query param to the endpoint and only fetch upw conditionally.

andymarke: Fair enough

PR #172, opened 2020-03-26
Fix so junit5 tests run, and integration tests that add to contact st…

andymarke:
…art from clean state


chris-faulkner-mojdt:
LGTM

PR #173, opened 2020-03-26
Pic 360 empty requirements

chris-faulkner-mojdt:
Changes to community API to return empty requirements . Also added unit / integration tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,6 +128,7 @@ task copyAgent(type: Copy) {
 assemble.dependsOn copyAgent
 
 test {
+    useJUnitPlatform()
     timeout = Duration.ofMinutes(15)
     useJUnitPlatform ()
     testLogging {
--- a/build.gradle
+++ b/build.gradle
@@ -111,8 +111,6 @@ dependencies {
     testImplementation 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
     testImplementation 'junit:junit:4.13'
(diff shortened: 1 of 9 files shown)


src/test/java/uk/gov/justice/digital/delius/service/ConvictionServiceTest.java
```diff
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.delius.service;
 
 import com.google.common.collect.ImmutableList;
+import org.junit.jupiter.api.Disabled;
```
andymarke: you can remove this now

build.gradle
```diff
@@ -111,6 +111,8 @@ dependencies {
     testImplementation 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
     testImplementation 'junit:junit:4.13'
+    testImplementation(platform("org.junit:junit-bom:5.6.0"))
+    testImplementation("org.junit.jupiter:junit-jupiter")
```
andymarke: I think this will be included already by spring boot test so it could be removed. Not sure what `bom` do you need it?

src/test/java/uk/gov/justice/digital/delius/service/ConvictionServiceTest.java
chris-faulkner-mojdt: Done

build.gradle
chris-faulkner-mojdt: Removed.

PR #174, opened 2020-04-01
DT-700 Endpoint to retrieve a user's authorised probation areas

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/UserService.java
```diff
@@ -119,6 +119,17 @@ public UserDetailsWrapper getUserDetailsList(final Set<String> usernames) {
                         .build());
     }
 
+    @Transactional
+    public Optional<UserAreas> getUserAreas(final String username) {
+        final var userWithAreas = userRepositoryWrapper.getUser(username);
```
andrewrlee, line 124: what happens when then doesn't find a user?

andymarke: That old wrapper throws a NotFoundException, so end up for a 404 - not sure why it was modelled using an Optional but since it ends in the same result I didn't want to refactor

PR #175, opened 2020-04-03
DT-644 Added sentence start date (need for comparing NOMIS sentences)

andymarke:


src/test/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformerTest.java
```diff
@@ -391,6 +391,17 @@ public void unpaidWorkIsNullWhereNonePresent() {
...
 
+    @Test
+    public void sentenceStartDateCopiedWhenPresent() {
+        Event event = Event.builder()
+                .disposal(Disposal.builder()
+                        .startDate(LocalDate.of(2020, 2, 22))
+                        .build())
+                .build();
+        final var conviction = transformer.convictionOf(event);
+        assertThat(conviction.getSentence().getStartDate()).isEqualTo(LocalDate.of(2020, 2, 22));
```
petergphillips, line 402: as a personal preference I never really like `LocalDate.of()` since I can never remember if the month starts from 0 or 1.  I find it clearer to do a `parse` with a date string instead

andymarke: Fair enough, I did some recent tests with parse in Kotlin with an extension function and it worked very well

PR #178, opened 2020-04-06
DT-645 Endpoint to update NOMS number

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -3,17 +3,40 @@
...
         log.info("NOMIS update NOMS number feature is {}", this.updateNomsNumberFeatureSwitch ? "ON" : "OFF");
     }
 
+    @Transactional
+    public IDs updateNomsNumber(String crn, UpdateOffenderNomsNumber updateOffenderNomsNumber) {
+        final var offender = offenderRepository.findByCrn(crn).orElseThrow(() -> new NotFoundException(String
+                .format("Offender with crn %s not found ", crn)));
+        if (updateNomsNumberFeatureSwitch) {
+            // TODO replace with real update
+            return offenderTransformer.idsOf(offender).toBuilder().nomsNumber(updateOffenderNomsNumber.getNomsNumber()).build();
```
andymarke, line 36: TODO - the meat to this service

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
```diff
@@ -5,10 +5,9 @@
 import org.springframework.http.MediaType;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.*;
-import uk.gov.justice.digital.delius.data.api.Custody;
-import uk.gov.justice.digital.delius.data.api.UpdateCustody;
-import uk.gov.justice.digital.delius.data.api.UpdateCustodyBookingNumber;
+import uk.gov.justice.digital.delius.data.api.*;
```
petergphillips, line 8: think it was decided that we shouldn't do wildcard imports - so I have changed my kotlin / java settings not to

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
```diff
@@ -56,5 +57,19 @@ public Custody updateCustodyBookingNumber(final @PathVariable String nomsNumber,
...
 
+    @RequestMapping(value = "offenders/crn/{crn}/nomsNumber", method = RequestMethod.PUT, consumes = "application/json")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "The offender record was updated", response = IDs.class),
+            @ApiResponse(code = 401, message = "Request is missing Authorization header (no JWT)"),
+            @ApiResponse(code = 404, message = "The requested offender was not found")
+    })
+    @ApiOperation(value = "Updates the offender record with the NOMS number in UpdateOffenderNomsNumber")
+    public IDs updateOffenderNomsNumber(final @PathVariable String crn,
+                                              final @RequestBody @Valid UpdateOffenderNomsNumber updateOffenderNomsNumber) {
```
petergphillips, line 68: does the `@Valid` annotation work without a `Validated` annotation? Can't see a test to confirm my suspicions...

src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -3,17 +3,40 @@
...
+        } else {
```
petergphillips, line 37: don't need an `else` since previous line has a `return`

src/test/java/uk/gov/justice/digital/delius/controller/secure/CustodyController_updateNomsNumberTest.java
```diff
@@ -0,0 +1,109 @@
...
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.*;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+public class CustodyController_updateNomsNumberTest {
+
+    private CustodyService custodyService = mock(CustodyService.class);
```
petergphillips, line 25: I thought in java land using the `mock` like this without the `RunWith` meant that the unnecessary stubbings etc. didn't run

petergphillips: although you're using rest assured so :man_shrugging:

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
petergphillips: actually just seen the 400 example below so sounds like it does, but wondering now about the `NotBlank` i.e. will spaces for the noms number work too?

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
andymarke: I think I changed my Kotlin stetting but not java, I'll correct my settings

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
andymarke: I'll add a test to check

src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
andymarke: interestingly I was looking at what the JDK does, and they do tend to `else` but agreed, change in next PR

PR #180, opened 2020-04-07
DT-645 NOMS number updated and SPG notified

andymarke:
Covers the basic update scenario where there are no duplicates etc

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
@@ -16,14 +16,17 @@ public class OffenderIdentifierService {
     private final boolean updateNomsNumberFeatureSwitch;
     private final OffenderTransformer offenderTransformer;
     private final OffenderRepository offenderRepository;
+    private final SpgNotificationService spgNotificationService;
 
     public OffenderIdentifierService(
             @Value("${features.noms.update.noms.number}") Boolean updateNomsNumberFeatureSwitch,
             OffenderTransformer offenderTransformer,
-            OffenderRepository offenderRepository) {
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/justice/digital/delius/service/SpgNotificationServiceTest.java
```diff
@@ -2,26 +2,45 @@
 
 import com.google.common.collect.ImmutableList;
 import lombok.val;
-import org.junit.Before;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.BeforeEach;
```
andymarke, line 5: moved to JUnit5 hence the changes

src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -16,14 +16,17 @@
     private final boolean updateNomsNumberFeatureSwitch;
     private final OffenderTransformer offenderTransformer;
     private final OffenderRepository offenderRepository;
+    private final SpgNotificationService spgNotificationService;
 
     public OffenderIdentifierService(
             @Value("${features.noms.update.noms.number}") Boolean updateNomsNumberFeatureSwitch,
```
matt-crowson-moj, line 22: An observation but would it pay to be clear with defaulting this to 'false'. `${features.noms.update.noms.number:false}`

And why object and not primitive? meh.

andymarke: Yeah, should be little `b` - not sure about default since it is default in application.yaml anyway and further defaulted in terraform. So it would probably just be a 3 level of defaults.

PR #183, opened 2020-04-08
DT-645 Fire SPG notification each time add AdditionalIdentifier

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -56,27 +56,37 @@ public IDs updateNomsNumber(String crn, UpdateOffenderNomsNumber updateOffenderN
...
+                            .build();
+                    offender.getAdditionalIdentifiers().add(additionalIdentifier);
+                    // force additional identifier to generate ID
+                    offenderRepository.flush();
+                    spgNotificationService
+                            .notifyInsertOfOffenderAdditionalIdentifier(offender, additionalIdentifier);
+                }
 
         );
-        offender.setNomsNumber(nomsNumber);
```
andymarke, line 79: move this above SPG message generation, so the SPG has the correct NOMS number

src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -56,27 +56,37 @@ public IDs updateNomsNumber(String crn, UpdateOffenderNomsNumber updateOffenderN
...
+                existingNomsNumber -> {
+                    final var additionalIdentifier = AdditionalIdentifier
+                            .builder()
+                            .identifier(existingNomsNumber)
+                            .offender(offender)
+                            .identifierName(referenceDataService.formerNomsNumberAdditionalIdentifier())
+                            .build();
+                    offender.getAdditionalIdentifiers().add(additionalIdentifier);
+                    // force additional identifier to generate ID
+                    offenderRepository.flush();
```
petergphillips, line 84: don't think `flush` will do anything since you're still in the same transaction which is only applied on commit?

andymarke: It should force the ID to be generated from the sequence, even if it isn't committed (which is all we need)

src/test/java/uk/gov/justice/digital/delius/service/SpgNotificationServiceTest.java
```diff
@@ -585,6 +586,107 @@ public void offenderIdIsSavedWithNotification() {
...
+                    Offender
+                            .builder()
+                            .offenderId(99L)
+                            .build(),
+                    AdditionalIdentifier
+                            .builder()
+                            .additionalIdentifierId(88L)
+                            .build());
+
+            verify(spgNotificationRepository, atLeastOnce()).saveAll(spgNotificationsCaptor.capture());
```
mikehalmamoj, line 609: I was just wondering why these tests all have `atLeastOnce()`... is it just that the test doesn't care about which areas have an interest?

PR #184, opened 2020-04-09
PIC-385 Add grouped documents call by CRN to OffendersResource and re…

chris-faulkner-mojdt:
Needed a "grouped" documents call in this API by CRN. Restored the unused one from OffenderController to OffenderResource. Added typing and some dates which we'll need in court case service.


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -120,7 +120,7 @@
                 offenderDetail -> new ResponseEntity<>(offenderDetail, OK)).orElse(new ResponseEntity<>(OffenderDetailSummary.builder().build(), NOT_FOUND));
     }
 
-    @ApiOperation(value = "Returns all document's meta data for an offender")
+    @ApiOperation(value = "Returns all document's meta data for an offender CRN")
```
andymarke, line 123: small typo, CRN -> NOMS Number

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderDocumentsByCrn.java
```diff
@@ -0,0 +1,113 @@
...
+                .auth()
+                .oauth2(validOauthToken)
+                .contentType(APPLICATION_JSON_VALUE)
+            .when()
+                .get(String.format(PATH_FORMAT, "X320741"))
+            .then()
+                .statusCode(HttpStatus.OK.value())
+                .extract()
+                .body()
+                .as(OffenderDocuments.class);
```
jhevans, line 92: I'm not suggesting you change anything here as it seems to be a favoured pattern here... but I'm interested in your thoughts. I'm not really a fan of this pattern for testing APIs where the JSON is marshalled back out to the response POJO. It assumes that the POJO itself is structured correctly and that it marshals to the correct JSON. If for example there's a typo in one of your POJO names this will miss it (assuming you use IDE auto completion which I think is probably a sensible assumption). I've seen this happen before in fact. At least if you use json paths you have to manually type in your path thus you'd have to make the error twice for it to pass in error.

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
chris-faulkner-mojdt: Done

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderDocumentsByCrn.java
andymarke: I think you are right, there is a risk (albeit small). 

Tests I do other projects tends to use JSON body assert methods instead. They sometimes can be ugly though.

I suppose best to judge risk v readability in the test

chris-faulkner-mojdt: I am not sure I agree.  I think it's too easy to forget to check fields when you are doing it like that. If someone adds a field, then they have to remember to update the JSON path. Also, it's not type safe. In the JSON I see this 

` .body("requirements[0].startDate", equalTo("2017-06-01"))`

I'd rather have the serialising technology make a date of it and check against that rather than a string.

I see this as being more readable. Also, if someone renames a field in Java then the test will continue to work. If they add a field, the test will fail because you are doing an object equality test. 

  ```
      final String getPath = String.format(PATH, CRN, SOME_CONVICTION_ID);
        final AttendancesResponse response = given()
            .accept(MediaType.APPLICATION_JSON_VALUE)
            .when()
            .get(getPath)
            .then()
            .statusCode(HttpStatus.OK.value())
            .extract()
            .body()
            .as(AttendancesResponse.class);

        assertThat(response.getCrn()).isEqualTo(CRN);
        assertThat(response.getConvictionId()).isEqualTo(SOME_CONVICTION_ID);
        assertThat(response.getAttendances()).hasSize(2);

        final AttendanceResponse expectedAttendance1 = AttendanceResponse.builder().contactId(1325L)
            .attendanceDate(LocalDate.of(2019, Month.AUGUST, 24))
            .attended(true).complied(true).outcome("Description of outcome")
            .contactType(ContactTypeDetail.builder().description("Description of contact type").code("DSC01").build()).build();
        final AttendanceResponse expectedAttendance2 = AttendanceResponse.builder().contactId(8888888L)
            .attendanceDate(LocalDate.of(2020, Month.FEBRUARY, 29))
            .attended(false).complied(false).outcome("8888888 Description of outcome")
            .contactType(ContactTypeDetail.builder().description("8888888 Description of contact type").code("DSC02").build()).build();

        assertThat(response.getAttendances()).containsExactlyInAnyOrder(expectedAttendance1, expectedAttendance2);
```

PR #186, opened 2020-04-17
Add an endpoint to retrieve the details of a single conviction

chris-faulkner-mojdt:
Code returns the same conviction type but just for a single ID.

Search by event ID which is PK on the EVENT table.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -334,7 +334,7 @@ public class OffendersResource {
     @ApiOperation(value = "Return the conviction (AKA Delius Event) for a conviction ID and a CRN")
     @ApiResponses(
         value = {
-            @ApiResponse(code = 200, message = "OK", response = Conviction.class, responseContainer = "List"),
+            @ApiResponse(code = 200, message = "OK", response = Conviction.class),
             @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
             @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
             @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -330,5 +330,28 @@ public class OffendersResource {
                 .map(convictionService::convictionsFor)
                 .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)));
     }
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -330,5 +330,28 @@ public OffenderDetail getOffenderDetailByNomsNumber(final @PathVariable("nomsNum
                 .map(convictionService::convictionsFor)
                 .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)));
     }
+
+    @ApiOperation(value = "Return the conviction (AKA Delius Event) for a conviction ID and a CRN")
+    @ApiResponses(
+        value = {
+            @ApiResponse(code = 200, message = "OK", response = Conviction.class, responseContainer = "List"),
```
andymarke, line 337: I guess `responseContainer` attribute isn't needed? Not sure what impact it has keeping it.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/EventRepository.java
```diff
@@ -10,6 +10,8 @@
 public interface EventRepository extends JpaRepository<Event, Long> {
     List<Event> findByOffenderId(Long offenderId);
 
+    List<Event> findByEventId(Long eventId);
```
andymarke, line 13: I think you can just use `findById` - I don't think you need  a new method, just use the one in the superclass. Which also means it will return an `Optional` rather than List which is what you want.

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -114,6 +114,17 @@ public ConvictionService(
                 .collect(toList());
     }
 
+    @Transactional(readOnly = true)
+    public Optional<Conviction> convictionFor(Long offenderId, Long eventId) {
+        List<uk.gov.justice.digital.delius.jpa.standard.entity.Event> events = eventRepository.findByEventId(eventId);
```
andymarke, line 119: once you use `findById` you can get rid of the `.stream()` and `.findFirst` below. And very minor prefer `final var` to typed left hand argument.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/EventRepository.java
chris-faulkner-mojdt: Thanks. Done.

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
chris-faulkner-mojdt: Yes

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
chris-faulkner-mojdt: Copy and paste from previous method. Yes.

PR #190, opened 2020-05-01
Pic 381 add breach

chris-faulkner-mojdt:
Added a general endpoint to fetch NSIs for an offender / conviction and an NSI type code. Intended to be used by court case service to just fetch breaches (limiting NSI type codes).

Decided to create an NsiService and NsiRepository because this is NSI focused. 

There was no NSI collection on the Event object , although there was a join from NSI to parent Event. I might have re-used conviction service to fetch events and search for the NSIs but did not want to add NSI to Event due to performance concerns. 

Added an NsiTransformer and removed the responsibilities for transforming NSI from the ContactTransformer. It looks like the existing transformers seem to think about using Spring by annotating with "Component" but then instantiate their own instances. There is a package visible requirementTransformer in ContactTransformer which is directly accessed by the AppointmentTransformer. Really thought about refactoring this out.


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -363,5 +364,29 @@ public Conviction getConvictionForOffenderByCrnAndConvictionId(
             .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)))
             .orElseThrow(() -> new NotFoundException(String.format("Conviction with ID %s for Offender with crn %s not found", convictionId, crn)));
     }
+
+    @ApiOperation(value = "Return the breaches for a conviction ID and a CRN")
+    @ApiResponses(
+        value = {
+            @ApiResponse(code = 200, message = "OK", response = Conviction.class),
```
andymarke, line 371: Little typo on response. In fact there has been a suggestion in DPS that we stop adding the 200 response annotation since apparently it will be generated automatically anyway, and always with the correct type, so may as well delete this line.

src/main/java/uk/gov/justice/digital/delius/data/api/NsiWrapper.java
```diff
@@ -0,0 +1,21 @@
...
+@ApiModel(description = "NSI Wrapper")
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class NsiWrapper {
+    @ApiModelProperty(value = "List of NSIs")
+    private List<Nsi> nsis;
+}
```
andymarke, line 21: it is so hard with initialism and plurals :-)

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.boot.web.server.LocalServerPort;
+import org.springframework.http.HttpStatus;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import uk.gov.justice.digital.delius.data.api.NsiWrapper;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("dev-seed")
+@RunWith(SpringJUnit4ClassRunner.class)
+public class OffendersResource_getOffenderNsisByCrn {
```
andymarke, line 26: This reminds me we still have all those tests disabled because of that memory issue, hopefully this one won't push it over the edge!

src/test/java/uk/gov/justice/digital/delius/service/NsiServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private NsiTransformer nsiTransformer;
+
+    @InjectMocks
+    private NsiService nsiService;
+
+    @DisplayName("All NSIs are fetched and returned, having been transformed")
+    @Test
+    void whenFetchNsisNoneFilterByCodeOrSoftDeleted() {
+
+        final uk.gov.justice.digital.delius.jpa.standard.entity.Nsi nsiEntity = buildNsi(EVENT, "BRE");
```
andymarke, line 45: very minor, but maybe use `var` instead of the types for readability

chris-faulkner-mojdt: Will do. Keep forgetting I am in Java 11 land :-)

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
chris-faulkner-mojdt: All OK so far in the pipeline.

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private static final String OFFENDERS_PATH = "/offenders/crn/%s/convictions/%s/nsis/%s";
```
jhevans, line 28: Is a path param an appropriate way of passing the NSI type? I'd expect this to be a queryParam as it's a filter?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
petergphillips: agreed, please remove the 200.  Worth also firing up the application and looking at the swagger docs to check they make sense e.g. duplicate `crn` documented below

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -363,5 +364,29 @@ public Conviction getConvictionForOffenderByCrnAndConvictionId(
...
+            @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
+            @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
+            @ApiResponse(code = 404, message = "The offender CRN is not found", response = ErrorResponse.class),
+            @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+        })
+    @GetMapping(path = "/offenders/crn/{crn}/convictions/{convictionId}/nsis/{nsiCodes}")
+    public NsiWrapper getNsiForOffenderByCrnAndConvictionId(
+        @ApiParam(name = "crn", value = "CRN for the offender", example = "A123456", required = true)
+        @NotNull @PathVariable(value = "crn") final String crn,
+        @ApiParam(name = "crn", value = "ID for the conviction / event", example = "2500295345", required = true)
```
petergphillips, line 382: duplicate `crn` field

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -363,5 +364,29 @@ public Conviction getConvictionForOffenderByCrnAndConvictionId(
...
+        @NotNull @PathVariable(value = "convictionId") final Long convictionId,
+        @ApiParam(name = "nsiCodes", value = "list of NSI codes to constrain by", example = "BRE,BRES", required = true)
+        @NotNull @PathVariable(value = "nsiCodes") final List<String> nsiCodes) {
```
petergphillips, line 385: this looks odd.  Filtering would normally done by query parameters rather than path variables, since you're getting a resource at `"/offenders/crn/{crn}/convictions/{convictionId}/nsis` and then filtering

petergphillips, line 385: should the `NotNull` actually be a `NotEmpty`?

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -0,0 +1,113 @@
...
+            .statusCode(HttpStatus.OK.value())
+            .extract()
+            .body()
+            .as(NsiWrapper.class);
+
+        assertThat(nsiWrapper.getNsis()).hasSize(1);
+    }
+
+    @Test
+    public void getGetOffenderNsisByCrnAndUnknownConvictionId_ReturnsEmptyCollection() {
```
jhevans, line 86: Further to the earlier comment about the NSI type possibly being a query param - I'd expect this to be a 404 if the conviction does not exist. 

For this URL`/offenders/crn/<some_crn>/convictions/<some_conviction_id>/nsis?nsiTypes=BRE,BRES` that would read to me something like "For the conviction with <some_conviction_id> belonging to the offender with CRN <some_crn> get me any NSIs with types in "BRE,BRES".

If either of the offender or the conviction does not exist that's a hard fail to my mind, but I don't know if there's some existing context I'm missing here?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
chris-faulkner-mojdt: OK - done

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -0,0 +1,113 @@
...
+        given()
+            .auth()
+            .oauth2(validOauthToken)
+            .contentType(APPLICATION_JSON_VALUE)
+            .when()
+            .get(String.format(OFFENDERS_PATH, "X777777", KNOWN_CONVICTION_ID, "BRES,BRE"))
+            .then()
+            .statusCode(HttpStatus.NOT_FOUND.value());
+    }
+}
```
jhevans, line 113: Could we add a test case for a known offender and conviction with no breach NSIs. i.e. returns an empty list

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
chris-faulkner-mojdt: Done

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
chris-faulkner-mojdt: Done

chris-faulkner-mojdt: PathVariable works and I believe offers better opportunities for cache control

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
andymarke: I think I agree, query param makes more sense

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
andymarke: Also agree, though in real world the conviction missing is probably "worse" than a 404 since it almost certainly means a coding error. So 404 is best but return empty list is probably not going to be too bad since we should never see a missing conviction, but may as well be consistent. Certainly shouldn't block the PR being merged.

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
petergphillips: cache control?  Have we got something in front of this doing the caching?

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
jhevans: As it happens there's another problem I've stumbled across here whilst I've been updating [PIC-389](https://dsdmoj.atlassian.net/browse/PIC-389) to bring it in line with the concept of an NSI rather than a breach as the resource. That would give us a clash between this and the url for a single NSI which would end up like this: `secure/offenders/crn/{crn}/convictions/{convictionId}/nsis/{nsiId}`

src/main/java/uk/gov/justice/digital/delius/transformers/ContactTransformer.java
```diff
@@ -35,7 +32,9 @@
 @Component
 public class ContactTransformer {
 
-    RequirementTransformer requirementTransformer = new RequirementTransformer();
+    final RequirementTransformer requirementTransformer = new RequirementTransformer();
```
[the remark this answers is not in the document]
chris-faulkner-mojdt: NO worries - I made it private and there's a compilation error. The AppointmentTransformer ought to really have its own instance of RequirementTransformer, or else a getter to retrieve the one from ContactTransformer. It's probably not worth doing much with this at the current time.

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
chris-faulkner-mojdt: Yes - agreed Query Param :-)

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
chris-faulkner-mojdt: So the first ever ticket  I did in community API we agreed that a missing conviction ID when fetching attendances would give en empty list. I was attempting to maintain consistency

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
chris-faulkner-mojdt: Will do.

PR #193, opened 2020-05-05
pic 381 - 404 if conviction ID doesn't 

chris-faulkner-mojdt:
Was returning 404 for a conviction ID which did exist for an offender in the events table but which had no entries in the NSI table. Wanted an empty list in this case.


src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -127,30 +148,30 @@ public void getGetOffenderNsisByCrnAndConvictionIdLimitCode() {
 
     @Test
     public void getGetOffenderNsisByCrnAndUnknownConvictionId_Returns404() {
```
andymarke, line 150: I'm confused how this test used to work, doesn't this show you did get a 404 for a unknown eventId?

andymarke: Ignore me, that wasn't the fix

PR #194, opened 2020-05-05
Pic 389 get nsi

jhevans:


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -410,5 +401,25 @@ public NsiWrapper getNsiForOffenderByCrnAndConvictionId(
...
+    @GetMapping(path = "/offenders/crn/{crn}/convictions/{convictionId}/nsis/{nsiId}")
+    public Nsi getNsiByNsiId(
+            @ApiParam(name = "crn", value = "CRN for the offender", example = "A123456", required = true)
+            @NotNull @PathVariable(value = "crn") final String crn,
+            @ApiParam(name = "convictionId", value = "ID for the conviction / event", example = "2500295345", required = true)
+            @NotNull @PathVariable(value = "convictionId") final Long convictionId,
+            @ApiParam(name = "nsiId", value = "ID for the nsi", example="123456", required = true)
+            @PathVariable(value = "nsiId") Long nsiId){
+        return offenderService.getOffenderByCrn(crn)
+                .map((offender) -> convictionService.convictionsFor(offender.getOffenderId())
```
andymarke, line 414: I wonder if we should just introduce a convictionService.findConvictionForOffender(long offenderId, long convictionId) since I suspect we will repeat this again and again

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/NsiRepository.java
```diff
@@ -1,14 +1,17 @@
...
 
+import java.util.List;
+
 public interface NsiRepository extends JpaRepository<Nsi, Long> {
 
     // There is no EVENT_ID on NSI
     @Query("select n from Nsi n where n.event.eventId = :eventId AND n.offenderId = :offenderId")
     List<Nsi> findByEventIdAndOffenderId(@Param("eventId") Long eventId, @Param("offenderId") Long offenderId);
+
+    Nsi getByNsiId(Long nsiId);
```
andymarke, line 16: do you need this, or can you use the super.findById() which returns an optional?

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
```diff
@@ -29,9 +54,25 @@ public NsiTransformer (@Autowired final RequirementTransformer requirementTransf
                 .actualStartDate(n.getActualStartDate())
                 .expectedStartDate(n.getExpectedStartDate())
                 .referralDate(n.getReferralDate())
+                .length(n.getLength())
+                .lengthUnit(NSI_LENGTH_UNIT)
```
andymarke, line 58: is this because the NSI doesn't have a unit so you have to hardcode?

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
```diff
@@ -29,9 +54,25 @@ public NsiTransformer (@Autowired final RequirementTransformer requirementTransf
...
+                .nsiManagers(nsiManagersOf(n.getNsiManagers()))
+                .court(courtTransformer.courtOf(nsi.getEvent().getCourt()))
```
andymarke, line 60: This feels odd, be interestingly to know why the frontend needs the event court (which I'm not sure what that is, the sentencing the court maybe??)

Maybe worth checking this and explaining the thinking around here it "just doesn't feel right"

src/main/java/uk/gov/justice/digital/delius/transformers/ProbationAreaTransformer.java
```diff
@@ -21,6 +21,10 @@ public ProbationAreaTransformer(InstitutionTransformer institutionTransformer) {
         this.institutionTransformer = institutionTransformer;
     }
 
+    public ProbationAreaTransformer() {
+        this.institutionTransformer = new InstitutionTransformer();
```
andymarke, line 25: Nice to see this, I hate all this Spring injection of transformers that is just a mess to test

src/test/java/uk/gov/justice/digital/delius/integration/secure/OffendersResource_getOffenderNsisByCrn.java
```diff
@@ -52,6 +58,76 @@ public void setup() {
                 new ObjectMapperConfig().jackson2ObjectMapperFactory((aClass, s) -> objectMapper));
     }
 
+    @Test
+    public void getNsiByCrnAndNsiId() {
+        String path = String.format(GET_NSI_PATH, KNOWN_CRN_FOR_NSI, KNOWN_CONVICTION_ID_FOR_NSI, KNOWN_NSI_ID);
+
+        Nsi nsi = given()
```
andymarke, line 65: very minor, but maybe use `final var`

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -25,16 +25,7 @@
...
-import uk.gov.justice.digital.delius.data.api.Contact;
-import uk.gov.justice.digital.delius.data.api.Conviction;
-import uk.gov.justice.digital.delius.data.api.CreatePrisonOffenderManager;
-import uk.gov.justice.digital.delius.data.api.NsiWrapper;
-import uk.gov.justice.digital.delius.data.api.OffenderDetail;
-import uk.gov.justice.digital.delius.data.api.OffenderDetailSummary;
-import uk.gov.justice.digital.delius.data.api.OffenderDocuments;
-import uk.gov.justice.digital.delius.data.api.OffenderLatestRecall;
-import uk.gov.justice.digital.delius.data.api.ResponsibleOfficer;
+import uk.gov.justice.digital.delius.data.api.*;
```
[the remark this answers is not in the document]
jhevans: It's never been something that's bothered me either way either 🤷‍♂️

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
jhevans: Yeah good idea, I think we already have a few instances. Will have a look

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
jhevans: Yeah, I did consider whether to exclude `lengthUnit` and just document instead but I know it does return a length and unit elsewhere so felt like this was more consistent and explicit. What do you reckon?

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
jhevans: Ok, I will try and get some clarification

src/main/java/uk/gov/justice/digital/delius/transformers/ProbationAreaTransformer.java
jhevans: Yeah I chewed this over for a bit as it was all getting very messy with dependencies getting passed down multiple levels. It produced a conflict between dependency inversion and the law of Demeter, strict dependency inversion meant passing ridiculous numbers of arguments, most of which were solely to be passed on in turn. I figured for testing we should only ever be concerned with immediate collaborators but we still need to be able to refactor over-sized transformers without breaking existing tests so thats where the default constructors come in.

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
andymarke: I think it is great idea, particularly that we have sentence lengths with units so good example how we can use APIs to improve the view of the data

src/main/java/uk/gov/justice/digital/delius/transformers/ProbationAreaTransformer.java
andymarke: Yeah, given transformers tend to be pure functions there is really no need to mock out. I did make a horrendous mistake where we have some Transformers for updates that do reference data lookups. I regret that now and need to find a better way

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Event.java
```diff
@@ -108,6 +108,10 @@
     @Column(name = "CPS_SOFT_DELETED")
     private Long cpsSoftDeleted;
 
+    @JoinColumn(name = "COURT_ID")
+    @OneToOne
+    private Court court;
```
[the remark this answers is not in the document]
jhevans: Yes I think you're right, also it's not an accurate representation of the relationship so should probably update it anyway. Will amend.

PR #197, opened 2020-05-12
DT-836 gradle plugin

mikehalmamoj:


.gitignore
```diff
@@ -63,4 +63,5 @@ dist
 !.elasticbeanstalk/*.global.yml
 
 # AWS DOcker build
-ecr.repo
\ No newline at end of file
+ecr.repo
+/dps-gradle-spring-boot-suppressions.xml
```
petergphillips, line 67: did you mean to add this here?  Shouldn't we be deleting `dependency-check-suppress-spring.xml` and using this instead?

build.gradle
```diff
@@ -1,119 +1,35 @@
...
-  suppressionFiles = ['dependency-check-suppress-spring.xml']
-  format = 'ALL'
-  analyzers {
-    assemblyEnabled = false
-  }
-}
-project.sourceCompatibility = 11
-project.targetCompatibility = 11
-
 group = 'uk.gov.justice.digital.delius'
```
petergphillips, line 6: be interesting to know whether this value has priority or whether the one from the plugin trumps it?

build.gradle
```diff
@@ -1,119 +1,35 @@
...
     implementation 'io.jsonwebtoken:jjwt:0.9.1'
-    implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
     implementation 'org.springframework.boot:spring-boot-starter-data-ldap'
-    implementation 'org.springframework.boot:spring-boot-starter-actuator'
     implementation 'org.springframework.boot:spring-boot-starter-security'
     implementation 'org.springframework.security.oauth:spring-security-oauth2:2.4.1.RELEASE'
     implementation 'org.springframework.security:spring-security-jwt:1.1.0.RELEASE'
 
     implementation 'org.hibernate:hibernate-java8:5.4.5.Final'
```
petergphillips, line 21: really old version of hibernate here - line can be removed as `data-jpa` will grab it instead.  Should be `hibernate-core` and then it will be updated to 5.4.15, but best removing.

build.gradle
mikehalmamoj: My understanding is that Gradle first applies the plugins then applies the buildscript - so I think it should override with this one :crossed_fingers:

build.gradle
```diff
@@ -1,119 +1,35 @@
...
-    implementation 'com.unboundid:unboundid-ldapsdk:4.0.13'
+    implementation 'com.unboundid:unboundid-ldapsdk:5.0.1'
```
andymarke, line 22: feels like a big bump? Is that deliberate?

.gitignore
andymarke: I think you still need to add that to gitignore though right? Else the build will keep adding it

build.gradle
mikehalmamoj: That was `useLatestVersions`... I'll put it back to the latest in the 4.X range.

.gitignore
petergphillips: yes, but we don't want it in `.gitignore`, we want to use the one from the plugin instead

.circleci/config.yml
```diff
@@ -25,25 +27,12 @@ jobs:
     executor: builder
     steps:
       - checkout
+      - hmpps/create_app_version
       - restore_cache:
           keys:
             - gradle-{{ checksum "build.gradle" }}
             - gradle-
       - run: ./gradlew build
```
petergphillips, line 35: think you'll need to pass the `APP_VERSION` in as `BUILD_NUMBER` here

.gitignore
mikehalmamoj: That is the one from the plugin (in gitignore)

.circleci/config.yml
```diff
@@ -66,11 +54,13 @@ jobs:
...
       - setup_remote_docker:
           docker_layer_caching: true
+      - mem/recall:
+          env_var: APP_VERSION
       - run:
           name: Build container and publish to docker hub
           command: |
-            APP_VERSION=$(cat ./app_version.txt)
             docker build -f docker/Dockerfile.aws --rm=false . \
+               --build-arg BUILD_NUMBER=$APP_VERSION \
```
petergphillips, line 63: if you do it this way then you'll need to change the Dockerfile to be a multi stage docker file instead - see auth project for example.  Also then don't need to do a `gradle build`, can do a `gradle test` instead

mikehalmamoj: Cool, thanks.  I wonder if it would be good to make the Gradle plugin generate the Dockerfile (and also run.sh) - would take more duplication out of the build.

.circleci/config.yml
```diff
@@ -25,25 +27,12 @@ jobs:
     executor: builder
     steps:
       - checkout
+      - hmpps/create_app_version
```
petergphillips, line 30: should be moved to the `build_docker` job

petergphillips: otherwise version when deploying is incorrect as I've just found out for auth

.circleci/config.yml
```diff
@@ -57,7 +46,6 @@ jobs:
           paths:
             - build/libs
```
petergphillips, line 47: don't need to save the paths any more

Dockerfile
```diff
@@ -1,3 +1,12 @@
+FROM openjdk:11-slim AS builder
+
+ARG BUILD_NUMBER
+ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
+
+WORKDIR /app
+ADD . .
+RUN ./gradlew clean assemble -Dorg.gradle.daemon=false
```
petergphillips, line 8: need to alter the `COPY` command below to copy from the builder

Dockerfile
```diff
@@ -13,7 +22,7 @@ RUN addgroup --gid 2000 --system appgroup && \
 
 WORKDIR /app
 
-COPY --chown=appuser:appgroup build/libs/community-api-*.jar /app/app.jar
+COPY --from=builder --chown=appuser:appgroup build/libs/community-api-*.jar /app/app.jar
 COPY --chown=appuser:appgroup build/libs/applicationinsights-agent*.jar /app/agent.jar
```
petergphillips, line 26: need to change all the `COPY` lines

PR #203, opened 2020-05-18
Fix: Prevent NullPointerExceptions for case where no disposal exists …

jhevans:
…for an event in RequirementService

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
@@ -6,12 +6,15 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import uk.gov.justice.digital.delius.controller.NotFoundException;
 import uk.gov.justice.digital.delius.data.api.ConvictionRequirements;
+import uk.gov.justice.digital.delius.jpa.standard.entity.Disposal;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -32,10 +35,14 @@ public ConvictionRequirements getRequirementsByConvictionId(String crn, Long con
                 .stream()
                 .filter(event -> convictionId.equals(event.getEventId()))
                 .map(Event::getDisposal)
-                .flatMap(disposal -> disposal.getRequirements().stream())
+                .flatMap(this::getRequirementStreamFromDisposal)
```
andymarke, line 38: This will work, as your test proves so let's merge this.
I suspect it is more complicated then needs be since the relationship between event to disposal is 1 -> (optional) 1

So I think semantically you probably wanted something like this with the key bit being the findAny() which shows there are not multiple events matching only one or none

`
        var requirements = eventRepository.findByOffenderId(offender.getOffenderId())
                .stream()
                .filter(event -> convictionId.equals(event.getEventId()))
                .filter(event -> event.getDisposal() != null)
                .findAny()
                .map(event -> event.getDisposal().getRequirements()
                        .stream()
                        .map(requirementTransformer::requirementOf)
                        .collect(Collectors.toList()))
                .orElse(List.of());
`

PR #205, opened 2020-05-19
DT-835 integration test source set

mikehalmamoj:
Move the integration tests to their own source set to isolate them which helps with future work to make them perform better.  A by product is that running `./gradlew test` no longer takes the rest of the day - to run both the unit tests and integration tests use command `./gradlew check`.


src/testIntegration/java/uk/gov/justice/digital/delius/OffenderHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.justice.digital.delius.jpa.standard.entity.OffenderAlias;
+import uk.gov.justice.digital.delius.jpa.standard.entity.OffenderManager;
+import uk.gov.justice.digital.delius.jpa.standard.entity.Officer;
+import uk.gov.justice.digital.delius.jpa.standard.entity.PartitionArea;
+import uk.gov.justice.digital.delius.jpa.standard.entity.ProbationArea;
+import uk.gov.justice.digital.delius.jpa.standard.entity.StandardReference;
+
+import java.time.LocalDate;
+
+public interface OffenderHelper {
```
mikehalmamoj, line 16: This is just moved, as are all test files.

src/testIntegration/java/uk/gov/justice/digital/delius/controller/api/DeliusOffenderAPITest.java
```diff
@@ -44,14 +54,13 @@
...
-import static uk.gov.justice.digital.delius.util.OffenderHelper.anOffender;
+import static uk.gov.justice.digital.delius.OffenderHelper.anOffender;
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, properties = {"offender.ids.pagesize=5"})
 @RunWith(SpringJUnit4ClassRunner.class)
 public class DeliusOffenderAPITest {
     @Rule
-    public WireMockRule wireMock = new WireMockRule(wireMockConfig().port(8088).jettyStopTimeout(10000L));
-
+    public WireMockRule wireMock = new WireMockRule(wireMockConfig().port(8088).usingFilesUnderDirectory("src/testIntegration/resources").jettyStopTimeout(10000L));
```
mikehalmamoj, line 63: A genuine change here - by default Wiremock gets its mappings from `src/test/resources`

build.gradle
```diff
@@ -1,12 +1,19 @@
 plugins {
     id("uk.gov.justice.hmpps.gradle-spring-boot") version "0.3.0"
     id 'groovy'
+    id 'org.unbroken-dome.test-sets' version '3.0.1'
```
mikehalmamoj, line 4: Plugin that configures additional source sets for testing.

src/testIntegration/java/uk/gov/justice/digital/delius/controller/api/DeliusOffenderAPITest.java
andymarke: So could we left the mapping files where they were, or do you think that was too confusing?

mikehalmamoj: I think it's not great that Wiremock hardcodes that mappings directory instead of looking on the classpath, so that confused me.  We could have left them where they were... doesn't feel good though.  I'll look a bit harder at using the classpath rather than trumping hardcoding with more hardcoding.

PR #207, opened 2020-05-19
DT-835 move to unit test

mikehalmamoj:
A couple of examples of replacing integration tests that mock out all beans with unit tests to test the logic and simple integration tests to test that the endpoint works e2e.  Also switched to Junit 5.

This was probably more effort than expected (though now I have the pattern down this should get quicker).

I think this change is valuable because:
* we now have some e2e tests which were previously missing
* the main logic is still covered, but by quicker unit tests
* JUnit 5 has forced an improvement in quality of mocks
* even if these endpoints are switched to `/secure` it should be straightforward to move the integration tests too
* we've got rid of the MockBeans so database is reused between integration tests

So the question is @andymarke do you think this is worth proceeding for the other 10 API classes I need to do the same for?

### DIFF
--- a/src/test/java/uk/gov/justice/digital/delius/controller/api/AppointmentControllerTest.java
+++ b/src/test/java/uk/gov/justice/digital/delius/controller/api/AppointmentControllerTest.java
@@ -28,7 +28,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
-public class AppointmentTest {
+public class AppointmentControllerTest {
 
     @Mock
     private AppointmentService appointmentService;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/justice/digital/delius/controller/api/AppointmentControllerTest.java
```diff
@@ -0,0 +1,298 @@
...
+                .then()
+                .statusCode(200)
+                .extract()
+                .body()
+                .as(Appointment[].class);
+
+        verify(appointmentService).appointmentsFor(eq(1L), appointmentFilterArgumentCaptor.capture());
+
+        assertThat(appointmentFilterArgumentCaptor.getValue().getAttended().orElse(null)).isEqualTo(Appointment.Attended.NOT_RECORDED);
+        assertThat(appointmentFilterArgumentCaptor.getValue().getFrom().orElse(null)).isEqualTo(LocalDate.of(2018, 11, 2));
```
andymarke, line 215: we would have preferred a json test, but lets not change that now - out of scope and the risk is so low

mikehalmamoj: Yeah ideally.  I'm a little worried about scope creep it's true - don't want this task to carry on forever - but when I'm done let's take another look at other such improvements.

PR #211, opened 2020-05-21
DT-858 Improve sorting and removed redundant id

andymarke:


src/main/java/uk/gov/justice/digital/delius/data/api/ReferenceData.java
```diff
@@ -11,8 +11,6 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class ReferenceData {
-    @ApiModelProperty(value = "Unique id for this reference data", example = "12345")
-    private String id;
```
andymarke, line 15: As suggested, no need to expose ID

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/ReferenceDataResourceTest.java
```diff
@@ -1,217 +0,0 @@
-package uk.gov.justice.digital.delius.controller.secure;
```
andymarke, line 1: Should have been deleted in earlier PR, these tests had already been merged into ReferenceDataAPITest

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesAPITest.java
```diff
@@ -73,12 +74,13 @@ public void setup() {
         jdbcTemplate.execute("DELETE FROM KEY_DATE");
         //noinspection SqlWithoutWhere
         jdbcTemplate.execute("DELETE FROM CONTACT");
+        flywayInstance = flyway;
     }
 
-    @After
```
andymarke, line 78: Ooops junit 4 annotation that was ignored

PR #217, opened 2020-05-26
DT-864 Convert Spring tests to pure unit tests and split out entity builders

andymarke:
Final part of transformer refactoring

1. Split a few more transformers between transforming a model object from an entity to the opposite, building a entity
2. Renamed (and moved) the class that build entities to <Name>EntityBuilder after realising these don't transform, they build
3 Any service test that previously needed the Spring Runner to inject the mess of transformers are now pure mocked tests
4 With two tests left that might benefit from using SpringRunner decided to make them consistent, and just use the `MockitoExtension`

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/entitybuilders/KeyDateEntityBuilder.java
+++ b/src/main/java/uk/gov/justice/digital/delius/entitybuilders/KeyDateEntityBuilder.java
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.delius.entitybuilders;
+
+import org.springframework.stereotype.Component;
+import uk.gov.justice.digital.delius.jpa.standard.entity.KeyDate;
+import uk.gov.justice.digital.delius.jpa.standard.entity.StandardReference;
+import uk.gov.justice.digital.delius.service.LookupSupplier;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+@Component
+public class KeyDateEntityBuilder {
+    private LookupSupplier lookupSupplier;
+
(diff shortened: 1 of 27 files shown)


src/main/java/uk/gov/justice/digital/delius/entitybuilders/KeyDateEntityBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Component
+public class KeyDateEntityBuilder {
+    private LookupSupplier lookupSupplier;
+
+    public KeyDateEntityBuilder(LookupSupplier lookupSupplier) {
+        this.lookupSupplier = lookupSupplier;
+    }
+
+    public KeyDate keyDateOf(uk.gov.justice.digital.delius.jpa.standard.entity.Custody custody, StandardReference keyDateType, LocalDate date) {
```
mikehalmamoj, line 19: I think you can import Custody now as it's the only one in this class?

src/test/java/uk/gov/justice/digital/delius/service/CourtReportServiceTest.java
```diff
@@ -21,22 +18,18 @@
...
 
-    @MockBean
-    private LookupSupplier lookupSupplier;
 
-
-
-    @Before
+    @BeforeEach
     public void before() {
+        courtReportService = new CourtReportService(courtReportRepository);
```
mikehalmamoj, line 32: I know it wasn't for this change but can we import CourtReport into this class to remove the full package references below?

PR #219, opened 2020-05-26
Dt 864 Flyway improvements and SpringBoot upgrade

andymarke:
1. Minor improvement to how we handle flyway restores
2. SpringBoot upgrade and fixes for breaking changes


src/main/resources/application-dev.yml
```diff
@@ -10,7 +10,3 @@ spring:
   h2:
     console:
       enabled: true
-
```
andymarke, line 13: default is off anyway, so removed

src/main/resources/application.yml
```diff
@@ -20,6 +20,8 @@ spring:
       enabled: true
   flyway:
     enabled: false
+  lifecycle:
+    timeout-per-shutdown-phase: 10s
```
andymarke, line 24: API requests are very quick

src/testIntegration/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepositoryTest.java
```diff
@@ -1,31 +0,0 @@
-package uk.gov.justice.digital.delius.jpa.standard.repository;
```
andymarke, line 1: No idea what the point of this test is, I think it loaded an offender in the old UKCloud database

src/main/java/uk/gov/justice/digital/delius/controller/advice/SecureControllerAdvice.java
```diff
@@ -109,4 +110,16 @@
                         .build());
     }
 
+    @ExceptionHandler(ResponseStatusException.class)
+    public ResponseEntity<ErrorResponse> handleException(final ResponseStatusException e) {
```
andymarke, line 114: Spring Boot upgrade changed the default behaviour for this exception, the message was no longer being propagated in the JSON.
So this is now consistent with other exception handlers

src/testIntegration/java/uk/gov/justice/digital/delius/FlywayRestoreExtension.java
```diff
@@ -0,0 +1,47 @@
...
+    public void afterAll(ExtensionContext extensionContext) {
+        logger.info("Cleaning database ..");
+        flyway.clean();
+        flyway.migrate();
+        logger.info("Cleaning database done");
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext extensionContext) {
+        final var applicationContext = getTestContextManager(extensionContext).getTestContext()
```
mikehalmamoj, line 39: SpringExtension.getApplicationContext is public and static, so could you just call that instead of copying from SpringExtension?

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesAPITest.java
```diff
@@ -36,7 +35,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-@ExtendWith(SpringExtension.class)
+@ExtendWith({SpringExtension.class, FlywayRestoreExtension.class})
```
mikehalmamoj, line 38: Very nice!

PR #222, opened 2020-06-04
DT-876 PR comments and use base test class

andymarke:
1. Minor PR comments for typo
2. Use base integration test hence remove a ton of duplicate code
3. Minor exception login change

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/advice/GeneralControllerAdvice.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/advice/GeneralControllerAdvice.java
@@ -28,7 +28,7 @@ public class GeneralControllerAdvice {
 
    @ExceptionHandler(NoSuchUserException.class)
    public ResponseEntity<String> noSuchUser(NoSuchUserException e) {
-       log.error("Unexpected exception", e);
+       log.info(e.getMessage());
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/advice/GeneralControllerAdvice.java
```diff
@@ -28,7 +28,7 @@
 
    @ExceptionHandler(NoSuchUserException.class)
    public ResponseEntity<String> noSuchUser(NoSuchUserException e) {
-       log.error("Unexpected exception", e);
+       log.info(e.getMessage());
```
andymarke, line 31: We get these for login failures, so just log as info not as an exception

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/AttendanceResourceAPITest.java
```diff
@@ -1,55 +1,26 @@
 package uk.gov.justice.digital.delius.controller.secure;
```
andymarke, line 2: All changes below is moving all integration tests to use the new base class and remove all the duplicated code for

1. Creating tokens
2. Configuring RestAssured
3. JSON writeObject stuff

PR #225, opened 2020-06-08
DT-876 Fix distinct issue by cutting down selected columns

andymarke:
Fix 2 issues:
1) select distinct in Oracle fails since it include all columns in OFFENDER including CLOBS
2) "APIUser" can't "see" 8 of the 2 millions offenders, so switching to "NationalUser" who should be able to see all offenders

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/OffenderPrimaryIdentifiers.java
+++ b/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/OffenderPrimaryIdentifiers.java
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.delius.jpa.standard.entity;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+import java.util.List;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder(toBuilder = true)
+@Entity
+@Table(name = "OFFENDER")
+public class OffenderPrimaryIdentifiers {
+    @Id
+    @Column(name = "OFFENDER_ID")
+    private Long offenderId;
+
+    @Column(name = "CRN")
+    private String crn;
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/OffenderPrimaryIdentifiers.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.persistence.Table;
+import java.util.List;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder(toBuilder = true)
+@Entity
+@Table(name = "OFFENDER")
+public class OffenderPrimaryIdentifiers {
```
andymarke, line 21: This means findAll will only select and distinct the few columns we are interested in, e.g exclude  CLOBs etc

src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -181,8 +184,10 @@ private Custody findCustodyOrThrow(Event activeCustodialEvent) {
                 .orElseThrow(() -> new CustodyNotFoundException(activeCustodialEvent));
     }
 
+    @NationalUserOverride
```
andymarke, line 187: This will use "NationalUser" rather than "APIUser". That user should have read rights on all offenders, APIUser seems to be missing 8 offenders out of the 2 million, possible because of the organisations associated with APIUser

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/OffenderPrimaryIdentifiers.java
```diff
@@ -0,0 +1,37 @@
...
+    @Column(name = "CRN")
+    private String crn;
+
+    @Column(name = "SOFT_DELETED")
+    private Long softDeleted;
+
+    @Column(name = "CURRENT_DISPOSAL")
+    private Long currentDisposal;
+
+    @OneToMany(mappedBy = "offenderId")
```
matt-crowson-moj, line 35: So just picking up on a conversation earlier.  Should this have the `FetchType.LAZY` specified?

andymarke: Not this one, it was the ManyToOnes that were the issue

PR #226, opened 2020-06-09
✨ pic 495 custodial status endpoint

jhevans:


src/test/java/uk/gov/justice/digital/delius/transformers/CustodialStatusTransformerTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertThat(custodialStatus.getSentenceDate()).isEqualTo(SENTENCE_DATE);
+        assertThat(custodialStatus.getActualReleaseDate()).isEqualTo(ACTUAL_RELEASE_DATE);
+        assertThat(custodialStatus.getLicenceExpiryDate()).isEqualTo(LICENCE_EXPIRY_DATE);
+        assertThat(custodialStatus.getLength()).isEqualTo(LENGTH);
+        assertThat(custodialStatus.getLengthUnit()).isEqualTo(LENGTH_UNIT);
+    }
+
+    @Test
+    public void givenMultipleReleases_thenActualReleaseDateIsTheLatest() {
+        Disposal disposal = buildDisposal(Arrays.asList(
```
jhevans, line 51: I've had to make an assumption here about what the correct behaviour is if there are multiple releases associated with a Disposal. I went for the simplest thing that seemed sensible but did consider whether it would be better to return the latest release that is not in the future? As it stands if a release is in the future it will nonetheless be returned but I'm not sure if a release can be in the future.

src/main/java/uk/gov/justice/digital/delius/data/api/CustodialTypeCode.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.delius.data.api;
+
+public enum CustodialTypeCode {
```
chris-faulkner-mojdt, line 3: Don't think this is used

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -31,27 +31,10 @@
...
-import uk.gov.justice.digital.delius.data.api.CommunityOrPrisonOffenderManager;
-import uk.gov.justice.digital.delius.data.api.Contact;
-import uk.gov.justice.digital.delius.data.api.Conviction;
-import uk.gov.justice.digital.delius.data.api.CreatePrisonOffenderManager;
-import uk.gov.justice.digital.delius.data.api.Nsi;
-import uk.gov.justice.digital.delius.data.api.NsiWrapper;
-import uk.gov.justice.digital.delius.data.api.OffenderDetail;
-import uk.gov.justice.digital.delius.data.api.OffenderDetailSummary;
-import uk.gov.justice.digital.delius.data.api.OffenderDocuments;
+import uk.gov.justice.digital.delius.data.api.*;
```
mikehalmamoj, line 34: So we had a discussion on wildcard imports or not on the DPS team and came to the conclusion they should be left in...  Has the probation team concluded differently?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -463,6 +447,29 @@ public Nsi getNsiByNsiId(
...
+                    @ApiResponse(code = 403, message = "Forbidden", response = ErrorResponse.class),
+                    @ApiResponse(code = 404, message = "The offender CRN is not found", response = ErrorResponse.class),
+                    @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+            })
+    @GetMapping(path = "/offenders/crn/{crn}/convictions/{convictionId}/sentences/{sentenceId}/custodialStatus")
+    public CustodialStatus getCustodialStatusBySentenceId(
+            @ApiParam(name = "crn", value = "CRN for the offender", example = "A123456", required = true)
+            @NotNull @PathVariable(value = "crn") final String crn,
+            @ApiParam(name = "convictionId", value = "ID for the conviction / event", example = "2500295345", required = true)
+            @NotNull @PathVariable(value = "convictionId") final Long convictionId,
```
mikehalmamoj, line 466: Do we need @NotNull on path variables?

src/main/java/uk/gov/justice/digital/delius/data/api/CustodialStatus.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.delius.data.api;
+
+import lombok.Builder;
+import lombok.Getter;
+
+import java.time.LocalDate;
+
+@Getter
+@Builder
+public class CustodialStatus {
```
mikehalmamoj, line 10: I know it's really hit and miss in the codebase, but do we need any @ApiModelProperty documentation on these?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
jhevans: Nope, Intellij did it automatically - we don't have a policy as such but I'll update my IDE settings and amend. Thanks for the heads up

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
jhevans: No don't see what effect it would have. Will remove.

src/main/java/uk/gov/justice/digital/delius/data/api/CustodialStatus.java
jhevans: Will review and add as appropriate, don't want to go crazy if names are self explanatory but there may well be non-obvious things that are worth documenting so good call.

src/main/java/uk/gov/justice/digital/delius/transformers/CustodialStatusTransformer.java
```diff
@@ -0,0 +1,59 @@
...
+import uk.gov.justice.digital.delius.data.api.CustodialStatus;
+import uk.gov.justice.digital.delius.data.api.KeyValue;
+import uk.gov.justice.digital.delius.jpa.standard.entity.Disposal;
+import uk.gov.justice.digital.delius.jpa.standard.entity.Release;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+@Component
+public class CustodialStatusTransformer {
```
mikehalmamoj, line 13: I think Andy recently switched most of the transformers to be static methods where it makes sense.  Mainly because we don't want to be mocking transformers anyway so there's no real value in making them beans.

src/main/java/uk/gov/justice/digital/delius/data/api/CustodialStatus.java
andymarke, line 10: So this is a sort of DTO, that takes stuff from the event , the disposal and custody, which is of course fine.
I suppose we could have returned a wrapper object that contained

- uk.gov.justice.digital.delius.data.api.Sentence
- uk.gov.justice.digital.delius.data.api.Custody 

some sort of CustodialSentence class. There is a danger we end up with a ton of different variations for each minor use case. Since we have duplicated the dates that were already in CustodyRelatedKeyDates for instance.

The thing missing if we did this is `offence` which is already in the conviction, those that indicate we are returning information the client has already from the /convictions endpoint?

Have a think if you think it is worth generalising this or if it is more trouble then it is worth. I'm not sure either way, just slightly worried that we might end up with two many endpoints that are too specific to very specific UIs

src/test/java/uk/gov/justice/digital/delius/service/SentenceServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+    public void whenCustodyIsSoftDeleted_thenReturnEmpty() {
+        when(offenderRepository.findByCrn(CRN)).thenReturn(Optional.of(offender));
+        when(disposalRepository.findByDisposalId(SENTENCE_ID)).thenReturn(Optional.of(disposal));
+        when(offender.getOffenderId()).thenReturn(OFFENDER_ID);
+        when(disposal.getOffenderId()).thenReturn(OFFENDER_ID);
+        when(disposal.getEvent()).thenReturn(event);
+        when(disposal.isSoftDeleted()).thenReturn(false);
+        when(disposal.getCustody()).thenReturn(custody);
+        when(custody.isSoftDeleted()).thenReturn(true);
+        Optional<CustodialStatus> status = sentenceService.getCustodialStatus(CRN, CONVICTION_ID, SENTENCE_ID);
```
mikehalmamoj, line 132: I rarely use the approach of mocking entities, but it's actually quite nice in that the only data mentioned is the data needed for the test - no test entity full of irrelevant data to reason about. :+1:

src/test/java/uk/gov/justice/digital/delius/transformers/CustodialStatusTransformerTest.java
andymarke: I very much doubt it, I suspect UI constrains it. If you are concerned we could always check database to see if there are any

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderCustodialStatusByCrnTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.time.LocalDate;
+
+import static io.restassured.RestAssured.given;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("dev-seed")
+public class OffendersResource_getOffenderCustodialStatusByCrnTest {
```
andymarke, line 25: You can now extend `IntegrationTestBase` and get rid of a lot of the setup code, see all the other refactored integration tests

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderCustodialStatusByCrnTest.java
```diff
@@ -0,0 +1,122 @@
...
+                .auth()
+                .oauth2(validOauthToken)
+                .contentType(APPLICATION_JSON_VALUE)
+                .when()
+                .get(path)
+                .then()
+                .statusCode(HttpStatus.OK.value())
+                .extract()
+                .body()
+                .as(CustodialStatus.class);
```
andymarke, line 63: I tend to now write JSON queries e.g;
`                .body("personalCircumstances[0].personalCircumstanceType.description", is("AP - Medication in Posession  - Assessment"));
`
I think I have come round to testing at JSON level rather than Model, gives a small advantage with asserting dates are correctly formatted

jhevans: Cool, that's still my preferred way of doing it as well. I went with this method for consistency but will feel free to go with JSON in future 👍

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderCustodialStatusByCrnTest.java
```diff
@@ -22,37 +14,20 @@
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @ActiveProfiles("dev-seed")
-public class OffendersResource_getOffenderCustodialStatusByCrnTest {
+public class OffendersResource_getOffenderCustodialStatusByCrnTest extends IntegrationTestBase{
```
mikehalmamoj: You can drop the annotations as they are inherited.

src/test/java/uk/gov/justice/digital/delius/service/SentenceServiceTest.java
andymarke: Personally I prefer starting from EntityBuilder.anOffender().toBuilder() etc and manipulating, but both work!

src/main/java/uk/gov/justice/digital/delius/service/SentenceService.java
```diff
@@ -0,0 +1,31 @@
...
+    public Optional<CustodialStatus> getCustodialStatus(String crn, Long convictionId, Long sentenceId) {
+        return offenderRepository.findByCrn(crn)
+            .map(offender -> disposalRepository.findByDisposalId(sentenceId)
+                    .filter(disposal -> offender.getOffenderId().equals(disposal.getOffenderId()))
+                    .filter(disposal -> convictionId.equals(disposal.getEvent().getEventId()))
+                    .filter(disposal -> !disposal.isSoftDeleted())
+                    .filter(disposal -> disposal.getCustody() != null)
+                    .filter(disposal -> !disposal.getCustody().isSoftDeleted())
+            )
+            .flatMap(optional ->  optional.map(custodialStatusTransformer::custodialStatusOf));
```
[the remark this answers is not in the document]
jhevans: Yeah that ought to work, I'll be honest I went with this method primarily due to convenience, familiarity and ease of testing. As much as it's a bit verbose it does read well imo, though so does the JPA equivalent. Given there's not a huge amount to choose between them I'll leave as is for now if that's ok?

PR #234, opened 2020-06-22
DT-926: Switch master to main and upgrade libs

petergphillips:


build.gradle
```diff
@@ -1,13 +1,11 @@
 plugins {
-    id("uk.gov.justice.hmpps.gradle-spring-boot") version "0.4.0"
+    id("uk.gov.justice.hmpps.gradle-spring-boot") version "0.4.2"
     id 'groovy'
     id 'org.unbroken-dome.test-sets' version '3.0.1'
 }
 
 group = 'uk.gov.justice.digital.delius'
 
-ext["spring-security.version"] = "5.3.2.RELEASE"
```
petergphillips, line 9: no need for pinning as latest spring boot contains latest spring security

src/test/java/uk/gov/justice/digital/delius/controller/secure/CustodyController_replaceNomsNumberTest.java
```diff
@@ -37,6 +39,7 @@
 
     @Before
     public void setup() {
+        RestAssuredMockMvc.config =  newConfig().encoderConfig(encoderConfig().defaultContentCharset("UTF-8"));
```
andymarke, line 42: Best solution I could find. It is connected to SpringBoot no longer defaulting charset which seems to only affect RestAssuredMockMVC (the RestAssured integration tests are fine, so it is only when RestAssuredMockMVC starts Spring we have the issue).

PR #235, opened 2020-06-22
DT-916 Change contract for replace NOMS number - when existing offend…

andymarke:
…er found return  400 rather then fail silently

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
@@ -86,7 +86,7 @@ public class CustodyResource {
 
     @RequestMapping(value = "offenders/nomsNumber/{originalNomsNumber}/nomsNumber", method = RequestMethod.PUT, consumes = "application/json")
     @ApiResponses(value = {
-            @ApiResponse(code = 400, message = "The new noms number is not present"),
+            @ApiResponse(code = 400, message = "The new noms number is not present in request or existing offender exists with then new NOMS number already"),
             @ApiResponse(code = 401, message = "Request is missing Authorization header (no JWT)"),
             @ApiResponse(code = 403, message = "Missing required role for this operation"),
             @ApiResponse(code = 404, message = "The requested offender was not found")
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
```diff
@@ -86,7 +86,7 @@ public IDs updateOffenderNomsNumber(final @PathVariable String crn,
 
     @RequestMapping(value = "offenders/nomsNumber/{originalNomsNumber}/nomsNumber", method = RequestMethod.PUT, consumes = "application/json")
     @ApiResponses(value = {
-            @ApiResponse(code = 400, message = "The new noms number is not present"),
+            @ApiResponse(code = 400, message = "The new noms number is not present in request or existing offender exists with then new NOMS number already"),
```
andrewrlee, line 89: could possibly model as a `409 Conflict`?

andymarke: @andrewrlee I like that a lot. @mikehalmamoj what do you think?

mikehalmamoj: :+1:

src/main/java/uk/gov/justice/digital/delius/service/OffenderIdentifierService.java
```diff
@@ -57,7 +58,7 @@ public IDs updateNomsNumber(String crn, UpdateOffenderNomsNumber updateOffenderN
 
         if (!existingOffendersAlreadyWithNomsNumber.isEmpty()) {
             log.info("No need to Update NOMS number since an offender with the new noms number already exists");
-            return existingOffendersAlreadyWithNomsNumber.stream().map(OffenderTransformer::idsOf).collect(Collectors.toList());
+            throw new ConflictingRequestException(String.format("NOMS number %s is already assigned to %s", updateOffenderNomsNumber.getNomsNumber(), existingOffendersAlreadyWithNomsNumber.stream().map(Offender::getCrn).collect(Collectors.joining(","))));
```
andymarke, line 61: Slightly archaic control by exception, but this follows existing pattern.  Something that we may want to move to Result type responses in the future ...

PR #236, opened 2020-06-26
AC-149 add endpoint to search for users by email address

tomsmyers:
In this pull request a new endpoint has been added to the `AuthenticationController` to return a list of Delius users with the supplied email address. The endpoint is currently accessible at `/secure/users/search/email/{email}/details`.

Behaviour:
* a list of users is always returned.
* if multiple users are found with the same email address, they are all returned.
* if no users are found, an empty list is returned (not 404).
* if a user is found in LDAP, but is missing from the delius internal user table, it is not included in the result.


src/main/java/uk/gov/justice/digital/delius/service/UserService.java
```diff
@@ -119,6 +119,38 @@ public UserDetailsWrapper getUserDetailsList(final Set<String> usernames) {
...
+                    .roles(user.getRoles().stream().map(role -> UserRole.builder().name(role.getCn()).build()).collect(toList()))
+                    .firstName(user.getGivenname())
+                    .surname(user.getSn())
+                    .email(user.getMail())
+                    .enabled(user.isEnabled())
+                    .userId(oracleUser.getUserId())
+                    .build();
+            })
+            .filter(user -> user != null)
+            .collect(Collectors.toList());
```
andymarke, line 151: I wonder if this would read slightly better if you extract a function?



```        return userList.stream()
            .map(user -> UserDetails.builder()
                    .roles(user.getRoles().stream().map(role -> UserRole.builder().name(role.getCn()).build()).collect(toList()))
                    .firstName(user.getGivenname())
                    .surname(user.getSn())
                    .email(user.getMail())
                    .enabled(user.isEnabled())
                    .userId(userIdOf(user.getCn()))
                    .build();
            )
            .filter(user -> user.getUserId() != null)
            .collect(Collectors.toList());`

private Long userIdOf(String cn) {
                try {
                    return userRepositoryWrapper.getUser(userCn).getUserId();
                } catch (NoSuchUserException e) {
                    log.error("no entry found in delius USER table for LDAP user '{}'", userCn);
                    return null;
                }

}
```

src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
```diff
@@ -65,6 +67,18 @@ public UserDetails findUser(@ApiParam(name = "username", value = "LDAP username"
...
     }
 
+    @ApiOperation(
+            value = "Find user details of a user held in Delius Identity (LDAP)")
+    @ApiResponses(
+            value = {
+                    @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+                    @ApiResponse(code = 401, message = "Unauthorised", response = ErrorResponse.class),
+            })
+    @RequestMapping(value = "/users/search/email/{email}/details", method = RequestMethod.GET)
```
andymarke, line 77: maybe `search` path is redundant

electrified:
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/AuthenticationController.java
tomsmyers: there was a reason i included this, but i'm happy to take it out.

the reason was to differentiate between the 'get user' style endpoints, which return either a user or 404, and a 'search' style endpoint, which returns a list of users (empty if not found). i figured perhaps in the future more 'search' style endpoints may be added for e.g. searching for a user using just part of the user name or a regex or something. e.g. `/users/search/{username}/details` which returns the same list style results.

src/main/java/uk/gov/justice/digital/delius/service/UserService.java
tomsmyers: like the idea, but i would want to reuse the `userIdOf` method in `getUserDetails` too, but in that case we don't want to swallow the exception. this makes me think the implementation is specific enough to just stay where it is.

PR #242, opened 2020-07-16
Add runbook and API docs badges to the readme

sldblog:
In case someone is interested in the [API documentation](https://community-api.test.delius.probation.hmpps.dsd.io/swagger-ui.html) but does not want to set up the project locally.

It seems the API docs is a recurring question on Slack, so maybe it's helpful to expose it directly.

The [runbook](https://dsdmoj.atlassian.net/wiki/spaces/NOM/pages/1739325587/DPS+Runbook#Deploying-Community-API) is also exposed as a badge, to make it easier to find.

The badges use https://shields.io and https://simpleicons.org.

image


philroffe:
I think this is a really good idea.  We have lots of good documentation but not necessarily easy to find (particularly for remote teams looking to integrate or check out integration capability).  In fact, it would be great to get the dev community view on this to see if we can adopt a bit of a standard set of badges/links?

mikehalmamoj:
Closing - the branch disappeared :man_shrugging:  Will do this on another branch.

sldblog:
I see it was merged in #274 🎉
Thank you @mikehalmamoj! 🙇‍♂️

PR #248, opened 2020-08-10
DT-1057 Increased max page size so that clients of getAllPrimaryIdent…

andymarke:
…ifiers can process huge pages that can be more efficient then lots of small ones.

### DIFF
--- a/src/testIntegration/resources/application.properties
+++ b/src/testIntegration/resources/application.properties
@@ -12,3 +12,5 @@ features.noms.update.keydates=true
 features.noms.update.noms.number=true
 
 spring.security.oauth2.resourceserver.jwt.public-key-location=classpath:local-public-key.pub
+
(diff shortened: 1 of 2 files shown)


src/testIntegration/resources/application.properties
```diff
@@ -11,4 +11,6 @@ features.noms.update.booking.number=true
 features.noms.update.keydates=true
 features.noms.update.noms.number=true
 
-spring.security.oauth2.resourceserver.jwt.public-key-location=classpath:local-public-key.pub
\ No newline at end of file
+spring.security.oauth2.resourceserver.jwt.public-key-location=classpath:local-public-key.pub
+
+logging.level.org.hibernate.SQL=DEBUG
```
andymarke, line 16: useful when running integration tests

PR #252, opened 2020-09-03
Pic 692 attendance data

chris-faulkner-mojdt:
After initial deployment, it turns out that our users need to view contacts / attendances based on other criteria. Current business analysis suggests it's a combination of the fields used.


andymarke (approved):
LGTM

PR #253, opened 2020-09-07
Pic 692 alter attendance sql

chris-faulkner-mojdt:
reviewed the SQL with Alistair - we need to return all contacts with an enforcement flag (can mean a null outcome type) OR not null contact outcome type so needed an outer join.

### DIFF
--- a/src/main/resources/db/data/V1_12__offender_add_contact_with_outcome_type_data.sql
+++ b/src/main/resources/db/data/V1_12__offender_add_contact_with_outcome_type_data.sql
@@ -1,0 +1,113 @@
+
+Insert into CONTACT (CONTACT_ID,
+                     CONTACT_DATE,
+                     OFFENDER_ID,
+                     NOTES,
+                     VISOR_CONTACT,
+                     STAFF_ID,
+                     TEAM_ID,
+                     SOFT_DELETED,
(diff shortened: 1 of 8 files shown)


src/main/resources/db/data/V1_12__offender_add_contact_with_outcome_type_data.sql
```diff
@@ -0,0 +1,113 @@
+
+Insert into CONTACT (CONTACT_ID,
```
jhevans, line 2: Just wondering whether this needs to be a separate migration or if it could be in with the rest of that offender's data. @andymarke what's best practice when adding new seed data? Should we be adding new incremental migrations or is it ok to just add it to the existing migrations? This test data is only ever used to seed transient in memory h2 dbs isn't it?

andymarke: So it just doesn't matter. I tend to amend the existing files, but sometimes create new files just to functionally separate the data.

The only real problem is there a couple of HUGE files that are so big that they don't even load in IntelliJ, so I tend to stay away from those.

andymarke (approved):
LGTM

PR #254, opened 2020-09-07
🐛 attendances "attended" and "complied" are represented by "Y" and "N…

chris-faulkner-mojdt:
Doing QA in dev / test environment, noticed that complied / attended were always returned as false. Turns out that the database stores these as Y / N. Seems like the database has various approaches to storage of what are boolean values.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/AttendanceService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/AttendanceService.java
@@ -4,6 +4,7 @@ import java.time.LocalDate;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/service/AttendanceService.java
```diff
@@ -35,7 +35,7 @@ public AttendanceService(final ContactRepository contactRepository) {
     static boolean forEntityBoolean(final String booleanStr) {
         if (booleanStr == null)
             return false;
-        return booleanStr.trim().equals("1");
+        return booleanStr.trim().equals("Y");
```
andymarke, line 38: There is TypesTransformer::ynToBoolean  if you prefer

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/service/AttendanceService.java
chris-faulkner-mojdt: Will use that. Thanks.

andymarke (approved):
LGTM

PR #255, opened 2020-09-08
DT-1099 Update keydates and POM now require update role

andymarke:
1. Endpoints that update POMs and Key Dates new require the ROLE_COMMUNITY_CUSTODY_UPDATE role
2. OWASP fix


petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyKeyDatesController.java
```diff
@@ -41,7 +41,8 @@ public CustodyKeyDatesController(OffenderService offenderService, ConvictionServ
             @ApiResponse(code = 400, message = "The keyDate is not valid or a key date can not be added to an offender which does not have a single custody event"),
             @ApiResponse(code = 404, message = "The requested offender was not found.")
     })
-    @ApiOperation(value = "Adds or replaces a custody key date for the active custodial conviction")
+    @ApiOperation(value = "Adds or replaces a custody key date for the active custodial conviction", notes = "Requires role ROLE_COMMUNITY_CUSTODY_UPDATE")
```
petergphillips, line 44: don't think `notes` does anything

PR #268, opened 2020-09-21
DT-1135 Wait more than 1 second before picking up offender updates to…

mikehalmamoj:
… avoid optimistic locking version clash


src/main/java/uk/gov/justice/digital/delius/service/OffenderDeltaService.java
```diff
@@ -48,11 +53,16 @@ public void deleteBefore(LocalDateTime dateTime) {
 
     @Transactional
     public Optional<OffenderUpdate> lockNextUpdate() {
-        final var mayBeDelta = offenderDeltaRepository.findFirstByStatusOrderByCreatedDateTime("CREATED");
+        final var createdCutOffTime = getCreatedCutOffTime();
```
andymarke, line 56: I think maybe not inline this

andymarke (approved):
LGTM

PR #277, opened 2020-10-08
DT-1184 Filter out custodial sentences in PSS for updates from DPS/OMiC

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -163,6 +163,7 @@ public Conviction addCourtCaseFor(Long offenderId, CourtCase courtCase) {
         val events = eventRepository.findByOffenderIdWithCustody(offenderId)
                 .stream()
                 .filter(Event::isActive)
+                .filter(event -> !event.getDisposal().getCustody().isPostSentenceSupervision())
```
mikehalmamoj, line 166: As discussed, it might be worth moving the active and postSentenceSupervision checks down into the repository layer?

mikehalmamoj (approved):

PR #282, opened 2020-10-13
DT-1203 Ignore deleted events when looking for custodial events

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -343,16 +337,10 @@ private String calculateNextEventNumber(Long offenderId) {
     }
 
     private List<Event> activeCustodyEvents(Long offenderId) {
-        return eventRepository
-                .findByOffenderId(offenderId)
+        return eventRepository.findByOffenderIdWithCustody(offenderId)
                 .stream()
-                .filter(event -> event.getSoftDeleted() == 0L)
-                .filter(event -> event.getActiveFlag() == 1L)
-                .filter(event -> event.getDisposal() != null)
```
andymarke, line 351: All these filters happen implicitly in the repo findBy

src/test/java/uk/gov/justice/digital/delius/jpa/standard/entity/EventTest.java
```diff
@@ -1,51 +1,88 @@
...
 
-    @Test
-    public void doesNotHaveCpsPackWhenDoesNotHaveAnAlfrescoId() {
-        assertThat(Event
-                .builder()
-                .cpsAlfrescoDocumentId("")
-                .cpsSoftDeleted(0L)
-                .build().hasCpsPack()).isFalse();
+        @Test
+        public void doesNotHaveCpsPackWhenHasAlfrescoIdAndNotDeletedTrue() {
```
lbennett-moj, line 32: should it be soft deleted in the test name?

lbennett-moj (approved):

src/test/java/uk/gov/justice/digital/delius/jpa/standard/entity/EventTest.java
andymarke: possibly, although I didn't change these test, just moved them inside a nested test class (which is no defence I suppose)

PR #285, opened 2020-10-16
Dt 1196 set responsible officer

andymarke:


src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_SwitchResponsibleOfficerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertThat(prisonOffenderManager(offenderManagersAfterSwitch).orElseThrow().getIsResponsibleOfficer()).isTrue();
+        assertThat(communityOffenderManager(offenderManagersAfterSwitch).orElseThrow().getIsResponsibleOfficer())
+                .isFalse();
+
+        // AND a CONTACT has been added to show responsible officer has changed
+        final var contacts = getRecentContacts("G0560UO");
+
+        assertThat(Arrays
+                .stream(contacts)
+                .anyMatch(contact -> contact.getContactType().getDescription().equals("Responsible Officer Change"))
```
mikehalmamoj, line 124: Should we check that we have 0 contacts before, then 1 contact after first RO switch, then 2 contacts after 2nd RO switch?

mikehalmamoj (approved):

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_SwitchResponsibleOfficerTest.java
andymarke: maybe, I opted not to, e.g if the first works not to bother to test again

PR #289, opened 2020-10-21
DT-1195 Get rid of confusing exceptions in log

andymarke:
Suppress exception under these categories that is polluting logs:

1. AWS `health/ping` is sending a bearer token that we don't care about
2. case notes correctly return 404 - but we don't need an exception in the logs for that


mikehalmamoj (approved):

chris-faulkner-mojdt (commented):
LGTM

PR #290, opened 2020-10-21
DT-1195 Allow retrieving when multiple registrations

andymarke:


src/main/java/uk/gov/justice/digital/delius/controller/advice/GeneralControllerAdvice.java
```diff
@@ -34,13 +36,15 @@
    }
 
     @ExceptionHandler(WebClientResponseException.class)
+    @Order(2)
```
petergphillips, line 39: that's quite nice having a default

petergphillips (dismissed):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/advice/GeneralControllerAdvice.java
mikehalmamoj: But why isn't the default the subclass?  I don't like it.

mikehalmamoj (approved):

PR #295, opened 2020-10-26
✨ PIC-830 add status date to NSI response

chris-faulkner-mojdt:
Unlike the other dates in this object (start, referral, actual start), there is a time component stored in the database. Don't want to throw away - consumers can disregard if they wish

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/Nsi.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/Nsi.java
@@ -19,7 +19,7 @@ public class Nsi {
     private KeyValue nsiSubType;
     private Requirement requirement;
     private KeyValue nsiStatus;
-    private LocalDateTime statusDate;
+    private LocalDateTime statusDateTime;
     private LocalDate actualStartDate;
     private LocalDate expectedStartDate;
     private LocalDate referralDate;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/Nsi.java
```diff
@@ -18,6 +19,7 @@
     private KeyValue nsiSubType;
     private Requirement requirement;
     private KeyValue nsiStatus;
+    private LocalDateTime statusDate;
```
andymarke, line 22: I don't have a strong opinion: do you think it should be called `statusDatetime` so it is obvious that this has a time element? Do we have a precedent for this when mixing types?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/Nsi.java
chris-faulkner-mojdt: I’ll rename the field. In AddressAssessment, there are 3 LocalDateTime fields, 2 have the suffix “Datetime” and one has “Date”. ApprovedPremisesReferral has a similar mix.

andymarke (approved):
LGTM

PR #301, opened 2020-11-12
DT-1308 Allow for multiple Staff in a prison with the same name. 

andymarke:
Problem:

When MOiC assigns a POM a new Staff member is created with same name as the POM. This Staff member is reused when a POM is assigned to a new prisoner.

This was error if more than one Staff member was found with the same name (JPA error).

This change chooses the first one, since MOiC only cares that the correct name is assigned, the Staff entity has no meaning other than the name.

### DIFF
--- a/src/test/java/uk/gov/justice/digital/delius/service/StaffServiceTest.java
+++ b/src/test/java/uk/gov/justice/digital/delius/service/StaffServiceTest.java
@@ -210,8 +210,8 @@ public class StaffServiceTest {
 
         List<StaffDetails> staffDetailsList = staffService.getStaffDetailsByUsernames(usernames);
 
-        var frazierUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("joefrazier")).findFirst().get();
-        var foremanUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("georgeforeman")).findFirst().get();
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/justice/digital/delius/service/StaffServiceTest.java
```diff
@@ -210,8 +210,8 @@ public void willReturnCorrectDetailsForMultipleUsernames_getStaffDetailsByUserna
 
         List<StaffDetails> staffDetailsList = staffService.getStaffDetailsByUsernames(usernames);
 
-        var frazierUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("joefrazier")).findFirst().get();
-        var foremanUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("georgeforeman")).findFirst().get();
+        var frazierUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("joefrazier")).findFirst().orElseThrow();
+        var foremanUserDetails = staffDetailsList.stream().filter(s -> s.getUsername().equals("georgeforeman")).findFirst().orElseThrow();
```
andymarke, line 214: unrelated IntelliJ warning fix

andrewrlee (approved):
LGTM

PR #306, opened 2020-11-18
Dt 1318 change prison location on pom allocate

andymarke:
Attempt to change the prison location when POM is allocated at a different prison.
If we are unable to change location that is fine.
TODO - telemetry for success and failure

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
+++ b/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
@@ -6,6 +6,7 @@ import org.hibernate.annotations.Where;
 import javax.persistence.*;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
```diff
@@ -242,4 +243,12 @@
                 .filter(PrisonOffenderManager::isActive)
                 .findFirst();
     }
+    public static class OffenderBuilder {
```
andymarke, line 246: Only used in tests that use builders(). Ensure POM list is mutable so we can amend that list

src/main/java/uk/gov/justice/digital/delius/service/OffenderManagerService.java
```diff
@@ -144,8 +144,10 @@ private CommunityOrPrisonOffenderManager allocatePrisonOffenderManager(final Pro
                         newPrisonOffenderManager.addResponsibleOfficer(responsibleOfficerRepository.save(responsibleOfficerOf(offender, newPrisonOffenderManager)));
                         contactService.addContactForResponsibleOfficerChange(newPrisonOffenderManager, existingPOM);
                     });
+            offender.getPrisonOffenderManagers().remove(existingPOM);
         }, () -> contactService.addContactForPOMAllocation(newPrisonOffenderManager));
 
+        offender.getPrisonOffenderManagers().add(newPrisonOffenderManager);
```
andymarke, line 150: ensure Offender is in the correct state for when JPA reuses it in a later transaction

mikehalmamoj (approved):

PR #308, opened 2020-11-20
DT-1334 Capitalise POMs names and find existing Staff regardless of case

andymarke:
OMiC will supply POM names in upper case; e.g JOHN SMITH

This resolves two issues:
1) When Staff already exists regardless of case - use that existing staff, so JOHN SMITH will match an existing john smith
2) When creating a new staff member capitalise - e.g transform to John Smith

### DIFF
--- a/src/main/resources/db/schema/V1_0__schema.sql
+++ b/src/main/resources/db/schema/V1_0__schema.sql
@@ -96,7 +96,7 @@ create sequence SC_PROVIDER_ID_SEQ
     ;
 
 create sequence STAFF_ID_SEQ
-    minvalue 2500000000
+    minvalue 2600000000
     maxvalue 900000000000000000
     cache 500
     ;
--- a/src/test/java/uk/gov/justice/digital/delius/service/OffenderManagerService_allocatePrisonOffenderManagerTest.java
+++ b/src/test/java/uk/gov/justice/digital/delius/service/OffenderManagerService_allocatePrisonOffenderManagerTest.java
@@ -2,6 +2,8 @@ package uk.gov.justice.digital.delius.service;
 
 import com.microsoft.applicationinsights.TelemetryClient;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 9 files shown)


src/main/resources/db/schema/V1_0__schema.sql
```diff
@@ -96,7 +96,7 @@ create sequence SC_PROVIDER_ID_SEQ
     ;
 
 create sequence STAFF_ID_SEQ
-    minvalue 2500000000
+    minvalue 2600000000
```
andymarke, line 99: boost index values in test data so new rows to have duplicate values

src/test/java/uk/gov/justice/digital/delius/service/OffenderManagerService_allocatePrisonOffenderManagerTest.java
```diff
@@ -80,502 +82,588 @@ public void setup() {
...
+        @Nested
+        @DisplayName("When successful")
+        class OnSuccess {
+            @BeforeEach
+            void setUp() {
+                when(staffService.findOrCreateStaffInArea(any(), any())).thenReturn(aStaff());
+            }
+
+            @Test
+            public void willCapitaliseStaffName() {
```
andymarke, line 648: NEW TEST HERE -->>>>

src/test/java/uk/gov/justice/digital/delius/service/OffenderManagerService_allocatePrisonOffenderManagerTest.java
```diff
@@ -80,502 +82,588 @@ public void setup() {
                 referenceDataService,
```
andymarke, line 82: I have only added one test here, the huge change is simple due to moving all tests into a `@Nested` class for two main service methods.  

Look for new test down the page.

chris-faulkner-mojdt (approved):
LGTM

PR #316, opened 2020-11-30
Dt 1360 making search by noms number case insensitive. 

dnyanesh-sonwane:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepository.java
```diff
@@ -25,12 +25,12 @@ public DuplicateOffenderException(String message) {
...
     Optional<Offender> findByCrn(String crn);
 
-    @Query("select o from Offender o where o.softDeleted = 0 and o.nomsNumber = :nomsNumber")
+    @Query("select o from Offender o where o.softDeleted = 0 and upper(o.nomsNumber) = upper(:nomsNumber)")
     Optional<Offender> findByNomsNumber(@Param("nomsNumber") String nomsNumber);
 
     // there are a small number of offenders (100 as of April 2020) that have duplicate NOMS numbers
     // this allows features that can deal with duplicates to access all offenders with the same number
-    @Query("select o from Offender o where o.softDeleted = 0 and o.nomsNumber = :nomsNumber")
+    @Query("select o from Offender o where o.softDeleted = 0 and upper(o.nomsNumber) = upper(:nomsNumber)")
```
chris-faulkner-mojdt, line 33: I think this is going to disable any indexing on noms number and mean that the table gets a full scan. I think in Oracle you can create an index based on the upper case version. 
Don't know the detail of NOMS numbers but is there a reason it isn't stored upper case in the first place ?

dnyanesh-sonwane: At the moment there is mix set of data. Most of them are in upper case, but  few in lower case.
Even if we update the storing part to store it in upper case, we will still have some existing data in lower case.

andymarke: What I suggest is we merge this and do a quick performance check in preprod. If we see an increase in times we back out

andymarke (approved):
LGTM

PR #320, opened 2020-12-04
✨ PIC-914 Add licence condition sub category

chris-faulkner-mojdt:


mikehalmamoj (approved):

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/LicenceCondition.java
```diff
@@ -55,6 +55,10 @@
     @OneToOne
     private LicenceConditionTypeMainCat licenceConditionTypeMainCat;
 
+    @JoinColumn(name = "LIC_COND_TYPE_SUB_CAT_ID")
+    @OneToOne
+    private StandardReference licenceConditionTypeSubCat;
```
andymarke, line 60: Interestingly this field is NULLABLE (but so is MainCat) - I have a feeling you might get an error or NullPointerException

andymarke (commented):
Is it possible to test this with null in the cat and sub cat fields?

PR #321, opened 2020-12-04
Pic 799 Restrict access to offender details based on exclusions and restrictions 


jhevans:
The full details and long term architectural plan for this can be found [here](https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/2043576390/Limited+Access+Markers+in+Community+API). In the first case this change should only affect the get offender by crn endpoints `/offenders/crn/X440890/all` and `/offenders/crn/X440890/`

Key changes:
- Introduced new roles ROLE_COMMUNITY_API_EXCLUDED_RESTRICTED and others - see doc above for full details. I've added these at the controller level as they should be non-breaking.
- Based on the client having these new roles the offender by crn endpoints will restrict access to users if they have been explicitly excluded or if the record is restricted and they are not on the relevant allow list.
- To this end have introduced a UserAccessService currently called from the controller which will throw an AccessDeniedException for the appropriate ControllerAdvice to handle returning a 403.


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -72,7 +74,7 @@
 @Slf4j
 @RequestMapping(value = "secure", produces = MediaType.APPLICATION_JSON_VALUE)
 @AllArgsConstructor
-@PreAuthorize("hasRole('ROLE_COMMUNITY')")
+@PreAuthorize("hasAnyRole('ROLE_COMMUNITY','ROLE_COMMUNITY_API_OPEN','ROLE_COMMUNITY_API_EXCLUDED','ROLE_COMMUNITY_API_EXCLUDED_RESTRICTED','ROLE_COMMUNITY_API_RESTRICTED')")
```
jhevans, line 77: Could apply this only on the relevant methods if that's preferable

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderByCrn.java
```diff
@@ -53,6 +54,56 @@ public void canGetOffenderSummaryByCrn() {
     assertThat(offenderDetail.getOtherIds().getCrn()).isEqualTo("X320741");
   }
 
+    @Test
+    public void givenUserIsNotExcluded_thenAccessAllowedWithAllRoles(){
+        final var username = "bernard.beaks";
+        final var path = "/offenders/crn/X440877";
+
```
jhevans, line 61: I went to town on the test cases here as it helped me get my head around the problem when I was starting out. For regression purposes it probably isn't needed as the complex logic is also tested at the UserAccessService level as well. Happy to pare these back if it's overkill.

andymarke (commented):
Looks like this has been implemented using the original ideas of roles. However we came to solution of using client credentials scopes works much better.

This has been implemented in probation-offender-search for the /phrase endpoint already.

Shall we have a chat which is the best way to go before merging this?

jhevans:
@andymarke I've updated as discussed, it simplified things a lot so pleased with that. This will potentially be breaking for any consumers of these endpoints that don't expect access control to be enforced though. Do we need to have a discussion to work out how we roll this out safely?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
andymarke: Do we still need this at all given we are using scope?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -282,17 +285,22 @@ public CommunityOrPrisonOffenderManager switchResponsibleOfficer(final @PathVari
...
+
+        return offenderService.getOffenderSummaryByCrn(crn)
+            .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)));
     }
 
     @RequestMapping(value = "/offenders/crn/{crn}/all", method = RequestMethod.GET)
     @ApiResponses(value = {
-            @ApiResponse(code = 404, message = "The offender is not found")
+            @ApiResponse(code = 404, message = "The offender is not found"),
+            @ApiResponse(code = 403, message = "Forbidden, the offender may have exclusions or restrictions in place preventing some users from viewing. Adopting client role ROLE_COMMUNITY_API_OPEN can bypass these restrictions."),
```
andymarke, line 298: does the comment need to be updated to explain it uses scope?

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -282,17 +285,22 @@ public CommunityOrPrisonOffenderManager switchResponsibleOfficer(final @PathVari
...
     })
     @ApiOperation(value = "Returns the full offender detail for the given crn", tags = "-- Popular core APIs --")
-    public OffenderDetail getOffenderDetailByCrn(final @PathVariable("crn") String crn) {
+    public OffenderDetail getOffenderDetailByCrn(final @PathVariable("crn") String crn, Authentication authentication) {
+        userAccessService.checkExclusionsAndRestrictions(crn, authentication.getAuthorities());
```
andymarke, line 302: Interestingly other projects does this slightly differently by injecting a SecurityUserContext where you can retrieve the Authentication, but this works just as well IMHO

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
```diff
@@ -0,0 +1,73 @@
...
+    private final OffenderService offenderService;
+    private final CurrentUserSupplier currentUserSupplier;
+    @Value("${user-access.scopes.dont-apply-exclusions-for}")
+    private final Set<String> ignoreExclusionRoles;
+    @Value("${user-access.scopes.dont-apply-restrictions-for}")
+    private final Set<String> ignoreRestrictionRoles;
+
+    public void checkExclusionsAndRestrictions(String crn, Collection<? extends GrantedAuthority> authorities) {
+        final var username = currentUserSupplier.username();
+        final var offender = offenderService.getOffenderByCrn(crn);
```
andymarke, line 27: I think you could move this down after we have established we need to do the check, else  I think you will be making the additional call when the scope says you don't need to check

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
```diff
@@ -0,0 +1,73 @@
...
+                .filter(AccessLimitation::isUserExcluded)
+                .map(accessLimitation -> new AccessDeniedException(accessLimitation.getExclusionMessage()));
+
+            if (excludedException.isPresent())
+                throw excludedException.get();
+        }
+
+        if (shouldCheckRestriction(authorities)) {
+            final var restrictedException = offender
+                .map(o -> username.map(u -> userService.accessLimitationOf(u, o))
```
andymarke, line 41: can't username be null here?

andymarke: ah, so username is an Optional ...

andymarke (approved):
This looks great, couple of minor comments about roles and swagger.

So if we have the default scopes (e.g no ignore aways type scope that no one will have right now except I think Pathfinder) then the following is true

With no user in scope we will get 403 for offender with a restricted list
With no user in scope we will get 200 for offender with a excluded list 

With a user in context then the list are honoured as expected

With the scopes that behaviour can be overridden. So to get back to the existing behaviour we need to add the scopes to all clients they use that API once this goes live

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
jhevans: Yeah I saw that and did consider updating it but clearly never got round to it happy to do it either way!

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
jhevans: I originally did have it like that but then having changed to the scoped configuration I figured our default case is that we will need to check so moved this up so it's only done once. I take your point that in that case it will be redundant though so will update.

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
jhevans: yeah, slightly funny, not hugely enamoured with the nested optionals but it was a sufficiently simple flow that I was happy with it. Could extract a `getAccessLimitationOf` method or similar if you think that would read better?

jhevans: Actually - better still I noticed there's precedent for using vavr.Either - think that might read better.

andymarke: Yeah, we have started using Either on new bits of code, I much prefer it to "flow by exception"

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
andymarke: One thing to remember is the biggest client of this API is the elastic search indexer that definitely won't have any restrictions. So that call would be redundant in the vast majority of cases.
One mitigation is the call *might* be cached by Hibernate - but not too sure.

jhevans: My reading on it suggests that it does but I couldn't swear to it. From what I understand it will cache for the DB session and I'd have thought that should be request scoped? It's all happening synchronously so seems unlikely that it wouldn't reuse the session anyway...

src/main/java/uk/gov/justice/digital/delius/service/UserAccessService.java
jhevans: Ignore my previous comment - I'm getting confused about how Either works!  Re using it in place of flow by exception I did consider doing something similar but then I ended up having to build my own responses in the controller and I figured that would add extra overhead for anyone wanting to reuse this code in future - and having the controller advice handle it seemed like a sufficiently idiomatic and well understood mechanism that I decided to keep it simple.

jhevans:
@andymarke could you take another look with the feature toggle enabled. I've defaulted to disabled but I'm not sure how to configure this on in the test environment?

andymarke (approved):
LGTM

PR #322, opened 2020-12-04
Pic 914 fix possible npe

chris-faulkner-mojdt:
Also added some unit test classes, covering some pre-existing classes


andymarke (approved):
LGTM

PR #325, opened 2020-12-15
Update test dependency and code to JUnit5

mxco86:
- Change the test dependency to JUnit5
- Update tests to use new dependency 
- Bit of wrangling to ensure Mockito's strictness settings are satisfied


andymarke (approved):
LGTM

PR #326, opened 2020-12-17
Dt 1423 pom allocation contact details

dnyanesh-sonwane:
Addition of email and phone to pom allocation api ( interface change only )


src/main/java/uk/gov/justice/digital/delius/data/api/ContactableHuman.java
```diff
@@ -0,0 +1,32 @@
...
+    @ApiModelProperty(value = "Email id", example = "officer@gov.uk")
+    private String email;
+    @ApiModelProperty(value = "Phone number", example = "0123411278")
+    private String phoneNumber;
+
+    public ContactableHuman capitalise() {
+        return this.toBuilder()
+                .surname(capitalizeFully(surname))
+                .forenames(capitalizeFully(forenames))
+                .email(capitalizeFully(email))
```
petergphillips, line 29: email should be lower case rather than capitalised

src/main/java/uk/gov/justice/digital/delius/data/api/ContactableHuman.java
```diff
@@ -0,0 +1,32 @@
...
+@Data
+@Builder(toBuilder = true)
+@NoArgsConstructor
+@AllArgsConstructor
+public class ContactableHuman {
+    @ApiModelProperty(value = "Given names", example = "Sheila Linda")    
+    private String forenames;
+    @ApiModelProperty(value = "Family name", example = "Hancock")    
+    private String surname;
+    @ApiModelProperty(value = "Email id", example = "officer@gov.uk")
```
petergphillips, line 20: would be better as `"Email address"` instead of `"Email id"`

src/main/java/uk/gov/justice/digital/delius/data/api/ContactableHuman.java
dnyanesh-sonwane: fixed

src/main/java/uk/gov/justice/digital/delius/data/api/ContactableHuman.java
dnyanesh-sonwane: fixed

petergphillips (approved):
LGTM

PR #327, opened 2020-12-18
✨ PIC-987 do not return soft deleted documents

chris-faulkner-mojdt:


src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/DocumentResourceAPITest.java
```diff
@@ -140,33 +140,7 @@ public void givenKnownCrnThenReturnDocuments() {
...
-                .filter(doc -> doc.getId().equals("1d842fce-ec2d-45dc-ac9a-748d3076ca6b")).findFirst().orElseThrow();
-            assertThat(courtReportDocument.getSubType().getCode()).isEqualTo("CJF");
-            assertThat(courtReportDocument.getSubType().getDescription()).isEqualTo("Pre-Sentence Report - Fast");
-            assertThat(courtReportDocument
-                .getReportDocumentDates()
-                .getRequestedDate()).isEqualTo(LocalDate.of(2018, 9, 4));
-            assertThat(courtReportDocument
-                .getReportDocumentDates()
-                .getRequiredDate()).isEqualTo(LocalDate.of(2019, 9, 4));
+            assertMultiDocuments(offenderDocuments);
```
chris-faulkner-mojdt, line 143: There were 25 lines duplicated here.

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/DocumentResourceAPITest.java
```diff
@@ -619,4 +567,41 @@ void withTypeAndSubTypeSuppliedOnlyDocumentsInThatTypeThatMatchTheSubTypeWillBeR
             }
         }
     }
+
+    private void assertMultiDocuments(OffenderDocuments offenderDocuments) {
+
+        assertThat(offenderDocuments.getDocuments()).hasSize(7);
+        assertThat(offenderDocuments.getDocuments()).extracting(OffenderDocumentDetail::getId)
+            .doesNotContain("4191easd-6e03-4bd8-b52d-9e050eefa745");
```
chris-faulkner-mojdt, line 575: This is the ID of the document I added in the new SQL. This line and the line which asserts all the IDs which should be present are now.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/CourtReportDocumentRepository.java
```diff
@@ -10,6 +10,6 @@
 import java.util.List;
 
 public interface CourtReportDocumentRepository extends JpaRepository<CourtReportDocument, Long>, JpaSpecificationExecutor<CourtReportDocument> {
-    @Query("select document from CourtReportDocument document, CourtReport entity where document.courtReport = entity and document.offenderId = :offenderId")
+    @Query("select document from CourtReportDocument document, CourtReport entity where document.courtReport = entity and document.offenderId = :offenderId and document.softDeleted = 0")
```
andymarke, line 13: Just missed a non-obvious change. Court reports also get retrieved using a Filter:
uk.gov.justice.digital.delius.jpa.filters.CourtReportDocumentFilterTransformer  (this is to support filtering of court reports, which is the exception).
So I think you just need to add something like 


`
        predicateBuilder.add(criteriaBuilder.equal(documentRoot.get("offenderId"), offenderId));
        **predicateBuilder.add(criteriaBuilder.equal(documentRoot.get("softDeleted"), 0));**

`

andymarke (dismissed):
LGTM - just see comment about that Filter wouldn't have been obvious

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/CourtReportDocumentRepository.java
chris-faulkner-mojdt: As far as I can see that filter is only used in the context of results returned from the repository in any case, in DocumentService. Not to say that we shouldn’t do it. It’s just that testing that change through DocumentServiceTest is odd because it means changing one of the repo mocks to return a soft deleted record which won't happen any more. 

Unfortunately there was no existing `CourtReportDocumentFilterTransformerTest` so will add one

andymarke (approved):
LGTM

PR #329, opened 2021-01-06
Dt 1439 prison location multiple custodial events

andymarke:
Previous if an offender had multiple sentences (events) we would not update the prison location on any of the events but return an error, this change will updated each active custodial event (based on current rules around custodial status)

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/Custody.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/Custody.java
@@ -6,6 +6,8 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.time.LocalDate;
+
 @Data
 @Builder(toBuilder = true)
 @NoArgsConstructor
@@ -20,4 +22,6 @@ public class Custody {
     private CustodyRelatedKeyDates keyDates;
     @ApiModelProperty(value = "Custodial status")
     private KeyValue status;
+    @ApiModelProperty(value = "Date when related sentence started")
+    private LocalDate sentenceStartDate;
 }
--- a/src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
@@ -96,7 +96,7 @@ public class CustodyService {
                 "toAgency", updateCustody.getNomsPrisonInstitutionCode());
 
         final var result = updateCustodyPrisonLocation(nomsNumber,
-                offender  -> getAllActiveConvictionByOffenderIdAndPrisonBookingNumber(offender.getOffenderId(), bookingNumber),
+                offender  -> getAllActiveConvictionsByOffenderIdAndPrisonBookingNumber(offender.getOffenderId(), bookingNumber),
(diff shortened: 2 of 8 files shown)


src/test/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformerTest.java
```diff
@@ -436,13 +435,6 @@ private Offence anOffence() {
                 .build();
     }
 
-    private uk.gov.justice.digital.delius.data.api.Offence anApiMainOffence() {
```
andymarke, line 439: no used

src/test/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformerTest.java
```diff
@@ -464,16 +456,8 @@ private Disposal aDisposal() {
                 .softDeleted(0L);
     }
 
-    private CourtCase aCourtCase() {
```
andymarke, line 467: not used

src/main/java/uk/gov/justice/digital/delius/data/api/Custody.java
```diff
@@ -20,4 +22,6 @@
     private CustodyRelatedKeyDates keyDates;
     @ApiModelProperty(value = "Custodial status")
     private KeyValue status;
+    @ApiModelProperty(value = "Date when related sentence started")
```
andymarke, line 25: added so custody objects can be distinguished when two sentences (events) combine to single time in prison

src/test/java/uk/gov/justice/digital/delius/service/CustodyServiceTest.java
```diff
@@ -413,17 +477,6 @@ public void willCreateTelemetryEventWhenConvictionNotFound() {
         }
 
 
-        @Test
-        @DisplayName("will add multiple events telemetry when more than one active custodial events found")
```
andymarke, line 417: moved (and amended) to nested class below

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -92,18 +96,21 @@ public Custody updateCustodyPrisonLocation(final String nomsNumber,
...
+                    telemetryClient.trackEvent("P2PTransferPrisonUpdated", add(telemetryProperties, "updatedCount", String.valueOf(success.custodyRecordsUpdated.size())), null);
                     break;
                 case NoUpdateRequired:
                     telemetryClient.trackEvent("P2PTransferPrisonUpdateIgnored", telemetryProperties, null);
                     break;
             }
-            return success.custody;
+            return success.custodyRecordsUpdated
+                .stream()
+                .max(Comparator.comparing(Custody::getSentenceStartDate))
```
mikehalmamoj, line 112: Just wondering why max?

mikehalmamoj (approved):

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
andymarke: We only return one of the custody records, so max essentially is the latest one which is most appropriate.

PR #335, opened 2021-01-11
DT-1424 Changes to store contact details for pom

dnyanesh-sonwane:


src/main/java/uk/gov/justice/digital/delius/transformers/StaffTransformer.java
```diff
@@ -40,10 +40,13 @@ static Human humanOf(Staff staff) {
                 .forenames(combinedMiddleNamesOf(staff.getForename(), staff.getForname2()))
                 .surname(staff.getSurname()).build();
     }
-    static ContactableHuman contactableHumanOf(Staff staff) {
+    static ContactableHuman contactableHumanOf(Staff staff, Optional<String> email, Optional<String> phoneNumber) {
```
andymarke, line 43: You could just make these Strings rather than Optional if that improves code noise

andymarke (changes requested):
LGTM.

I think you change the status of a PR so it can't be merged, I forgotten how.

.circleci/config.yml
```diff
@@ -68,7 +68,6 @@ workflows:
             branches:
               only:
                 - main
-
```
andymarke, line 71: Is this a mistake?

src/main/resources/db/schema/V1_0__schema.sql
```diff
@@ -9050,6 +9050,8 @@ create table PRISON_OFFENDER_MANAGER
     CREATED_BY_USER_ID         NUMBER           not null,
     LAST_UPDATED_DATETIME      DATE             not null,
     LAST_UPDATED_USER_ID       NUMBER           not null,
+    EMAIL_ADDRESS VARCHAR2(255),
```
andymarke, line 9053: I wonder if it would be better to comment these out?? At least it proves everything works with out those columns

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_AllocatePrisonOffenderManagerTest.java
```diff
@@ -446,6 +446,48 @@ void willReturnAConflictWhenSupplyingANOMSNumberWithDuplicateActiveOffenders() {
                 .statusCode(409);
     }
 
+    //TODO - Enable this test when EMAIL_ADDRESS and TELEPHONE_NUMBER columns are added in production
+    //@Test
```
andymarke, line 450: Vey minor, I think `@Disabled`  might have been better so you can give a nice description why is not run

andymarke (approved):
LGTM

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_AllocatePrisonOffenderManagerTest.java
dnyanesh-sonwane: Done

.circleci/config.yml
dnyanesh-sonwane: Fixed

src/main/resources/db/schema/V1_0__schema.sql
dnyanesh-sonwane: Done

andymarke (approved):
LGTM

PR #340, opened 2021-01-15
DT-1503 Added more telemetry so we can capture removed key dates

andymarke:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -337,7 +387,7 @@ private void addContactForBulkCustodyKeyDateUpdate(Long offenderId, Event event,
         }
     }
 
-    private Map<String, LocalDate> datesAmendedOrUpdated(Map<String, LocalDate> keyDatesToBeAddedOrUpdated) {
+    private Map<String, LocalDate> datesAddedOrUpdated(Map<String, LocalDate> keyDatesToBeAddedOrUpdated) {
```
andymarke, line 390: misnamed; unconnected to change

src/test/java/uk/gov/justice/digital/delius/service/ConvictionService_AddOrReplaceOrDeleteCustodyKeyDatesTest.java
```diff
@@ -241,4 +245,138 @@ void willDoNothingWhenNoDatesHaveChanged() {
         verify(eventRepository, never()).save(any());
     }
 
+    @Nested
+    @MockitoSettings(strictness = Strictness.LENIENT)
```
mikehalmamoj, line 249: Surprised you have to set this explicitly - I thought that was the default?

mikehalmamoj (approved):

src/test/java/uk/gov/justice/digital/delius/service/ConvictionService_AddOrReplaceOrDeleteCustodyKeyDatesTest.java
andymarke: It would appear not.

TBH I think it is a sign of a test not structured correctly In this case the parent test assumes it will was aways update a key date hence the need to ignore services never called.

PR #345, opened 2021-01-22
PIC-674 Add path ending "custodialStatus" to allow compatibility 

chris-faulkner-mojdt:
Will remove the endpoint after court case service has migrated forward


andymarke (approved):
LGTM

PR #355, opened 2021-02-01
MAS-296 Add Sentence Type to Convictions Endpoints

markreesmoj:


src/main/java/uk/gov/justice/digital/delius/data/api/Sentence.java
```diff
@@ -30,4 +30,6 @@
     private LocalDate startDate;
     private LocalDate terminationDate;
     private String terminationReason;
+    @ApiModelProperty(value = "Type of sentence")
+    private String sentenceType;
```
andymarke, line 34: I would have thought maybe a KeyValue might be better so you get the description and code rather than just the code

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/Sentence.java
markreesmoj: Good call, I have done so

andymarke (approved):
LGTM

PR #358, opened 2021-02-02
DT-1542 implement MAPPA details service

mikehalmamoj:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/RegistrationRepository.java
```diff
@@ -1,10 +1,21 @@
...
 public interface RegistrationRepository extends JpaRepository<Registration, Long> {
     List<Registration> findByOffenderId(Long offenderId);
+
+    @Query("select registration from Registration registration " +
+        "where registration.registerType.code = 'MAPP' " +
+        "and registration.offenderId = :offenderId " +
+        "and registration.softDeleted = 0 " +
+        "and registration.deregistered = 0 " +
+        "order by registration.createdDatetime desc")
+    Page<Registration> findActiveMappaRegistrationByOffenderId(Long offenderId, Pageable pageable);
```
mikehalmamoj, line 20: As JPQL lacks a `limit` and we only want the first result, implemented pageable so we can just get a single page with a single element - effectively `limit 1`.

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/service/RiskService.java
```diff
@@ -1,12 +1,25 @@
...
+    }
+
     public MappaDetails getMappaDetails(Long offenderId) {
-        return null;
+        final var reg = registrationRepository.findActiveMappaRegistrationByOffenderId(offenderId, PageRequest.of(0, 1));
+            return reg.stream()
+                .findFirst()
+                .map(registration -> MappaDetailsTransformer.mappaDetailsOf(registration))
+                .orElseThrow(() -> new NotFoundException("MAPPA details for offender not found"));
     }
```
andymarke, line 24: I can't see an alternative to a 404 but I wonder in longer term if we need a clear way of saying what entity is not found, something like

`"reason": "offender"`
or 
`"reason": "mappa"`

Maybe we need some feedback from clients since I have always found this possibly a cause for confusion.
Having said that I suspect the fact a client is making this call they have already established the offender exists so maybe it doesn't matter.

Either nothing to do for this change

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/service/RiskService.java
mikehalmamoj: I know what you mean, but I think it's a rare occasion that the client cares why - I haven't come across it much.  So yes I think we only do that if the client asks for it.

PR #362, opened 2021-02-09
MAS-283 update offender tier

markreesmoj:
hmpps-tier will write the recalculated tier back into delius via community-api
A new role has been created for this new endpoint, but it's just a string in this project at the moment. What else has to happen to allow it to be used?

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -372,11 +372,12 @@ public class OffendersResource {
             .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)));
     }
 
-    @ApiOperation(value = "Update offender tier", tags = {"Assessments"})
+    @ApiOperation(value = "Update offender tier. Requires ROLE_MANAGEMENT_TIER_UPDATE", tags = {"Assessments"})
     @ApiResponses(
         value = {
             @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
-            @ApiResponse(code = 404, message = "The offender CRN is not found", response = ErrorResponse.class),
+            @ApiResponse(code = 403, message = "Forbidden, requires ROLE_MANAGEMENT_TIER_UPDATE"),
+            @ApiResponse(code = 404, message = "The offender CRN or Tier is not found", response = ErrorResponse.class),
             @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
         })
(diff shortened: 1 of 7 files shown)


src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_updateTier.java
```diff
@@ -0,0 +1,79 @@
...
+            .oauth2(tokenWithRoleManagementTierUpdate())
+            .contentType(APPLICATION_JSON_VALUE)
+            .when()
+            .post("/offenders/crn/X320741/tier/B1")
+            .then()
+            .statusCode(200).extract()
+            .body()
+            .as(OffenderDetail.class);
+
+        assertThat(updatedTierOffender.getCurrentTier()).isEqualTo("B1");
```
andymarke, line 39: I think maybe also do a another GET like above, since that proves it was definitely persisted and it was your POST just echoing back what was sent

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -371,6 +372,23 @@ public OffenderAssessments getAssessmentsByCrn(
...
+            @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+            @ApiResponse(code = 404, message = "The offender CRN is not found", response = ErrorResponse.class),
+            @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+        })
+    @PostMapping(path = "/offenders/crn/{crn}/tier/{tier}")
+    @PreAuthorize("hasRole('ROLE_MANAGEMENT_TIER_UPDATE')")
+    public OffenderDetail updateTier(
+        @ApiParam(value = "CRN for the offender", example = "A123456", required = true)
+        @NotNull @PathVariable(value = "crn") final String crn,
+        @ApiParam(value = "New tier", example = "A1", required = true)
```
andymarke, line 387: maybe document the list of tier codes (even though it is reference data it does mean the docs offer a quick glance at the possible values)

andymarke (dismissed):
LGTM

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_updateTier.java
markreesmoj: Done

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
markreesmoj: Done

andymarke (approved):
LGTM

PR #364, opened 2021-02-09
pat-1636 Add 'unallocated' ldus to probation area and ldu endpoint. S…

lbennett-moj:
…ome ldus that are not 'selectable' in delius have offender manager/offender allocation


andymarke (approved):
LGTM

PR #365, opened 2021-02-12
DT-1613 extract jwt scopes

mikehalmamoj:
Fixes a bug where the JWT scopes were not being copied into authorities and so didn't work.


petergphillips (approved):
LGTM

PR #366, opened 2021-02-12
DT-1616 Cater for null mappa level and category

mikehalmamoj:


src/main/java/uk/gov/justice/digital/delius/transformers/MappaDetailsTransformer.java
```diff
@@ -30,6 +35,26 @@ public static MappaDetails mappaDetailsOf(Registration registration) {
 
     }
 
+    private static String getCategoryDescription(Optional<StandardReference> maybeCategory) {
```
mikehalmamoj, line 38: I know Optional isn't a great idea on a method parameter, but it makes the transformer much easier to read :man_shrugging:

andymarke (approved):
LGTM

PR #372, opened 2021-02-25
MAS-317 Update tier

markreesmoj:
composite key on MANAGEMENT_TIER


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -383,12 +385,13 @@ public OffenderAssessments getAssessmentsByCrn(
         })
     @PostMapping(path = "/offenders/crn/{crn}/tier/{tier}")
     @PreAuthorize("hasRole('ROLE_MANAGEMENT_TIER_UPDATE')")
-    public OffenderDetail updateTier(
+    public ResponseEntity updateTier(
```
andymarke, line 388: Very minor, I wonder if this was `void` it would just work and return 200 with no body??

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ManagementTierId.java
```diff
@@ -0,0 +1,30 @@
...
+public class ManagementTierId implements Serializable {
+    @Column(name = "OFFENDER_ID")
+    private Long offenderId;
+
+    @ManyToOne
+    @JoinColumn(name = "TIER_ID")
+    private StandardReference tier;
+
+    @Column(name = "DATE_CHANGED")
+    private LocalDateTime dateChanged;
```
andymarke, line 29: So I guess these are never updated, we just add new ones as they change? And Delius uses the latest?

src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
```diff
@@ -208,22 +206,5 @@ private Custody findCustodyOrThrow(Event activeCustodialEvent) {
                         .build());
     }
 
-    @Transactional
```
andymarke, line 211: Nice to see this move to its own service 👍🏿

src/main/java/uk/gov/justice/digital/delius/service/TierService.java
```diff
@@ -0,0 +1,68 @@
...
+        Map<String, String> telemetryProperties = Map.of("crn", crn, "tier", tier);
+        final Long offenderId = getOffender(crn, telemetryProperties);
+
+        writeTierUpdate(tier, telemetryProperties, offenderId);
+
+        telemetryClient.trackEvent("TierUpdateSuccess", telemetryProperties, null);
+    }
+
+    private Long getOffender(String crn, Map<String, String> telemetryProperties) {
+        return offenderRepository.findByCrn(crn).map(Offender::getOffenderId).orElseThrow(() -> logAndThrow(telemetryProperties, "TierUpdateFailureOffenderNotFound", String.format("Offender with CRN %s not found", crn)));
```
andymarke, line 42: Nice, I like the idea of the telemetry on exception 👍🏿

src/main/java/uk/gov/justice/digital/delius/service/TierService.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private void writeTierUpdate(String tier, Map<String, String> telemetryProperties, Long offenderId) {
+        final StandardReference updatedTier = standardReferenceRepository.findByCodeAndCodeSetName(String.format("U%s", tier), "TIER").orElseThrow(() -> logAndThrow(telemetryProperties, "TierUpdateFailureTierNotFound", String.format("Tier %s not found", tier)));
+
+        final StandardReference changeReason = standardReferenceRepository.findByCodeAndCodeSetName("ATS", "TIER CHANGE REASON").orElseThrow(() -> logAndThrow(telemetryProperties, "TierUpdateFailureTierChangeReasonNotFound", "Tier change reason ATS not found"));
```
andymarke, line 48: Maybe move this into ReferenceDataService so we have one place for all these codes

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ManagementTierId.java
markreesmoj: No, there aren't any updates to this table as I understand it. It looks like it provides a history of tier changes. 

I am not clear what Delius does downstream. All the values we're writing are at the instruction of Michael Wetherall. 

This table is pretty odd - no primary key column and half the columns are never used

src/main/java/uk/gov/justice/digital/delius/service/OffenderService.java
markreesmoj: It's going to grow like a triffid in the next ticket unfortunately

src/main/java/uk/gov/justice/digital/delius/service/TierService.java
markreesmoj: Makes sense

PR #373, opened 2021-03-01
pat-1664 Fix to return recall reason on the Recall entity replacing i…

lbennett-moj:
…ncorrect mapping to Standard reference

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Recall.java
+++ b/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Recall.java
@@ -5,7 +5,11 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Recall.java
```diff
@@ -5,7 +5,11 @@
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import javax.persistence.*;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
 import java.time.LocalDateTime;
```
lbennett-moj, line 13: is the standard still individual imports?

andymarke (approved):
LGTM

PR #375, opened 2021-03-03
MAS-323 create an SPG Notification

markreesmoj:
There isn't an integration test for this work because SpgNotificationHelperNoProceduresRepositoryImpl which is present in the profile used by the tests never returns any interested CRCs, hence no notifications will be created


andymarke (approved):
LGTM

PR #383, opened 2021-03-22
DAPI-128 DAPI-130 Move Delius specific config from interventions and Use sentence id for creating NSI

mahalrmoj:
Provides sentenceId to community-api as part of the ReferralSentRequest so it can used when creating a NSI in Delius, and builds on DAPI-128:

Moves the responsibility of the assignment of these NSI codes (on ReferralSent) to community-api.

1. provider-code: CRS
2. staff-code: CRSUATU
3. team-code: CRSUAT
4. Referral type notion is replaced by service category

A NSI Referral in Delius requires a sentenceId, building on DAPI-128 which ensures all Delius specific codes and transformations are isolated to and are the concern of community api.


src/main/resources/application.yml
```diff
@@ -100,6 +100,19 @@ features:
...
+  referral:
+    provider-code: CRS
+    staff-code: CRSUATU
+    team-code: CRSUAT
+    nsi-status: INPROG
+    referral-type-to-nsi-type: '{
+      "Accommodation": "CRS01",
+      "Education, Training and Employment (ETE)": "CRS02",
+      "Finance, Benefits and Debt (FBD)": "CRS03",
+      "Dependency and Recovery (D&R)": "CRS04"
```
electrified, line 113: CRS05 - CRS07 are also in the reference data that's been set up for Interventions

src/main/resources/application.yml
```diff
@@ -100,6 +100,19 @@ features:
         update-prison-location: false
   apply-limited-access-markers: false
 
+new-nsi:
+  referral:
+    provider-code: CRS
+    staff-code: CRSUATU
+    team-code: CRSUAT
+    nsi-status: INPROG
+    referral-type-to-nsi-type: '{
```
electrified, line 109: This could be in the standard YAML map format?

electrified (approved):

src/main/resources/application.yml
mahalrmoj: They haven't been included as there is no direct mapping from service categories in Interventions to CRS05-07. This is under review and the current thinking is service category will probably be replaced with Contract Title. However, this work will feature as part of another PR.

src/main/resources/application.yml
mahalrmoj: Because of the spaces and special characters, the standard YAML format becomes a little unreadable.

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
```diff
@@ -12,24 +12,43 @@
...
 
     private final NsiService nsiService;
 
     private final OffenderService offenderService;
 
-    public ReferralService(final DeliusApiClient deliusApiClient, final NsiService nsiService, final OffenderService offenderService) {
+    public ReferralService(final DeliusApiClient deliusApiClient,
+                           final NsiService nsiService,
+                           final OffenderService offenderService,
+                           @Value("${new-nsi.referral.provider-code}") final String providerCode,
```
andymarke, line 38: You might want to group these into a config object using @ConfigurationProperties(prefix = "new-nsi")

andymarke (approved):
LGTM

src/main/resources/application.yml
mahalrmoj: Found that if I contain the name in square brackets it's fine. Will change:

    service-category-to-nsi-type:
      "[Accommodation]": CRS01
      "[Education, Training and Employment (ETE)]": CRS02
      "[Finance, Benefits and Debt (FBD)]": CRS03
      "[Dependency and Recovery (D&R)]": CRS04

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
mahalrmoj: Will do

andymarke (approved):
LGTM

andymarke (approved):
LGTM

electrified (approved):

src/main/resources/application.yml
```diff
@@ -100,6 +100,18 @@ features:
...
+  nsi-mapping:
+    provider-code: CRS
+    staff-code: CRSUATU
+    team-code: CRSUAT
+    nsi-status: INPROG
+    service-category-to-nsi-type:
+      "[Accommodation]": CRS01
+      "[Education, Training and Employment (ETE)]": CRS02
+      "[Finance, Benefits and Debt (FBD)]": CRS03
+      "[Dependency and Recovery (D&R)]": CRS04
```
sldblog, line 113: 🙋‍♂️ To note, the service category titles might change without a second thought; their UUIDs won't.

That said, I don't think this is worth changing now as we will have "contract type" mapping in the near future.

mxco86: +1 for mapping using the UUID -> Delius ref. We'll include this in the next ticket

PR #387, opened 2021-03-25
DAPI-129 Deriving requirement id from sentence and requirement type

mahalrmoj:
This change allows community-api to derive the "requirement" for which an offender (referral) NSI will be created. It is derived for a specific sentence and a requirement type of Rehabilitation Activity (RAR).  

This change allows clients like interventions-service to only have to identify and keep track of a sentence for an offender and not need any knowledge of requirements that exist in Delius.


src/main/resources/application.yml
```diff
@@ -112,6 +112,10 @@ interventions-delius:
       "[Finance, Benefits and Debt (FBD)]": CRS03
       "[Dependency and Recovery (D&R)]": CRS04
 
+delius-nsi:
+  requirement:
+    rehabilitation-activity-type: F
+
```
andymarke, line 118: Very minor (so leave it as is), but IMHO since this is not environment specific not sure it gains from being in the config when it could just be a hardcoded constant, just more levels of indirection when trying to understand what is going on.

andymarke (approved):
LGTM

electrified (approved):

PR #388, opened 2021-03-30
MAS-365 offer the option to retrieve active convictions only

markreesmoj:


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -144,7 +144,7 @@
             @NotNull @PathVariable(value = "nomsNumber") final String nomsNumber) {
```
andymarke, line 144: Could you do the same here for requests by nomsNumbers to make it consistent. I think it is an important feature that we should have for prison based services as well

andymarke (dismissed):
LGTM

andymarke (approved):
LGTM

PR #389, opened 2021-03-30
Get only the conviction we want

markreesmoj:
Spotted this when I had to change this method when changing the convictions service for MAS-365, and it looks like it should just retrieve a single conviction instead of getting them all then filtering on convictionId.


andymarke (approved):
LGTM

PR #390, opened 2021-03-31
FIX-DEP new version of spring boot gradle plugin to fix dependency check

markreesmoj:


andymarke (approved):
you could do 3.1.4 if you want for few minor extras

PR #392, opened 2021-04-01
ALS-2724 Enabled deployments to ECS

marcus-bcl:
Uses the [aws-ecs](https://circleci.com/developer/orbs/orb/circleci/aws-ecs) orb, with AWS credentials stored in CircleCI contexts, to enable deployment to the various Delius environments.

This enables automatic deployment to the `delius-core-dev` environment on every git push, to enable better feedback against pull requests. This may be contentious so feel free to remove it if it's a problem.

Screenshot 2021-04-01 at 13 43 10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,10 @@ version: 2.1
 
 orbs:
   hmpps: ministryofjustice/hmpps@3.0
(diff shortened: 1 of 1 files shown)


andymarke (approved):
LGTM

PR #395, opened 2021-04-08
DT-1570 Resourcing risk allocation endpoint so that MPC can allocated…

andymarke:
This adds a new endpoint to retrieve the latest case allocation for an offender.
Pre-"Probation Day 1" this would indicate if the person is to be managed by NPS or CRC, in the post re-nationalisation world this will indicate the resourcing allocation level - e.g. whether a high/medium risk person  requires more resourcing.

This endpoint returns that decision.
This has been requested by MPC since it affects how prison offender managers are allocated to cases.


src/main/java/uk/gov/justice/digital/delius/transformers/CaseAllocationTransformer.java
```diff
@@ -0,0 +1,35 @@
...
+import uk.gov.justice.digital.delius.jpa.standard.entity.CaseAllocation;
+
+import java.util.Optional;
+
+public class CaseAllocationTransformer {
+    private static final String RETAINED_CODE = "R";
+
+    public static RiskResourcingDetails riskResourcingDetailsOf(CaseAllocation caseAllocation) {
+        final var resourcingDecision = Optional
+            .ofNullable(caseAllocation.getAllocationDecision())
```
andymarke, line 14: database suggests this can be null

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/RiskAPITest.java
```diff
@@ -55,102 +59,255 @@ void missingRole_forbidden(String uri) {
 
 
     @Nested
-    class NotFound {
+    class Mappa {
+        @Nested
+        class NotFound {
```
andymarke, line 65: just moved inside nested test

karenmillermoj (approved):
lgtm

PR #399, opened 2021-04-09
DAPI-88 Improvements/Changes in Interventions-CommunityApi request and responses as a result of E2E Integration 

mahalrmoj:
After completing the end to end integration between Interventions and Delius UI (via community-api and delius-api) a number of improvements and changes were identified in order to have a working integration. These comprise:

1. Provide offsets with dates and times so that appointments and referrals can be exchanged correctly with Delius
2. Provide integration context (applicable config) in all requests
3. Use UUIDs over long text to identify reference data (i.e. service category) 

To achieve a working integration between new Interventions Service and the existing Case Management System Delius.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateRequest.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateRequest.java
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.delius.data.api;
 
-import com.fasterxml.jackson.annotation.JsonFormat;
 import io.swagger.annotations.ApiModelProperty;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
@@ -8,8 +7,7 @@ import lombok.Data;
 import lombok.NoArgsConstructor;
 
 import javax.validation.constraints.NotNull;
-import java.time.LocalDate;
-import java.time.LocalTime;
+import java.time.OffsetDateTime;
 
 @Data
 @Builder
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateRequest.java
```diff
@@ -19,18 +17,11 @@
 
     @NotNull
     @ApiModelProperty(required = true)
-    @JsonFormat(pattern="yyyy-MM-dd")
-    private LocalDate appointmentDate;
```
andymarke, line 23: What happened here? Is Delius in some odd timezone or something?

src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+
+public class DateConverter {
+
+    private static final String LOCAL_TIMEZONE = "Europe/London";
```
andymarke, line 11: I think you could do system timezone since we are always London Europe as define in Docker file

src/main/resources/application.yml
```diff
@@ -110,10 +110,10 @@ delius-integration-context:
       nsi-mapping:
         nsi-status: INPROG
         service-category-to-nsi-type:
-          "[Accommodation]": CRS01
-          "[Education, Training and Employment (ETE)]": CRS02
-          "[Finance, Benefits and Debt (FBD)]": CRS03
-          "[Dependency and Recovery (D&R)]": CRS04
+          428ee70f-3001-4399-95a6-ad25eaaede16: CRS01
```
andymarke, line 113: I'm a bit confused why you need these GUIDs can't you just use the right hand code?

andymarke (commented):
LGTM, some questions

src/main/resources/application.yml
mahalrmoj: We are translating from interventions service categories to Delius and it was thought that the description field was too fragile and the only handle remaining was the UUID.

andymarke: So those the calling service passthrough something like`428ee70f-3001-4399-95a6-ad25eaaede16` in the API request?

mahalrmoj:
This PR is being withdrawn as Delius-api is being changed to provide dates and times as OffSetDateTimes. This means there will be no ambiguity about what they represent and is in line with MoJ standards. Once done, this PR will be redone.

PR #401, opened 2021-04-13
MAS-341 remove fallback for absence of tiering reference data

markreesmoj:
The reference data exists in all environments now following the deployment of the weekend's delius release PDM 21.10


andymarke (approved):
LGTM

PR #403, opened 2021-04-13
✨ PIC-1288 flag to include teams in the ProbationAreas 

chris-faulkner-mojdt:
Flag defaults to false. Observation on this and other similar endpoints is that the inclusion of all teams on probation areas can massively increase the size of the JSON. Consumers of these endpoints do not use them.


andymarke (approved):
LGTM

PR #404, opened 2021-04-13
DAPI-88 Interventions Capi Integration Improvements

mahalrmoj:

After completing the end to end integration between Interventions and Delius UI (via community-api and delius-api) a number of improvements were identified in order to have a working integration. These comprise:

- Interventions to provide offsets with dates and times so that appointments and referrals can be converted correctly for Delius which relies on Europe/London local dates and times
- Use UUIDs over long text to identify reference data (e.g. service category). The use of descriptions was deemed fragile ( business codes are not available for service categories)
- Provide integration context in all requests to identify the correct mapping configuration for a client of community-api

To achieve a workable integration between new Interventions Service and the existing Case Management System Delius for both referrals and appointments creation.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
+++ b/src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
@@ -1,0 +1,24 @@
+package uk.gov.justice.digital.delius.utils;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+
+public class DateConverter {
+
+    private static final String LOCAL_TIMEZONE = "Europe/London";
+
+    public static LocalDate toLondonLocalDate(OffsetDateTime offsetDateTime) {
+        return offsetDateTime.atZoneSameInstant(ZoneId.of(LOCAL_TIMEZONE)).toLocalDate();
+    }
+
+    public static LocalDateTime toLondonLocalDateTime(OffsetDateTime offsetDateTime) {
+        return offsetDateTime.atZoneSameInstant(ZoneId.of(LOCAL_TIMEZONE)).toLocalDateTime();
+    }
+
+    public static LocalTime toLondonLocalTime(OffsetDateTime offsetDateTime) {
+        return offsetDateTime.atZoneSameInstant(ZoneId.of(LOCAL_TIMEZONE)).toLocalTime();
+    }
+}
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -110,10 +110,10 @@ delius-integration-context:
       nsi-mapping:
         nsi-status: INPROG
         service-category-to-nsi-type:
-          "[Accommodation]": CRS01
-          "[Education, Training and Employment (ETE)]": CRS02
-          "[Finance, Benefits and Debt (FBD)]": CRS03
-          "[Dependency and Recovery (D&R)]": CRS04
+          428ee70f-3001-4399-95a6-ad25eaaede16: CRS01
+          ca374ac3-84eb-4b91-bea7-9005398f426f: CRS02
+          96a63c39-4371-4f17-a6ec-265755f0cf7b: CRS03
+          76bcdb97-1dea-41c1-a4f8-899d88e5d679: CRS04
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+
+public class DateConverter {
+
+    private static final String LOCAL_TIMEZONE = "Europe/London";
```
mxco86, line 11: Maybe add a comment here to indicate why the timezone is hard-coded?

mxco86 (approved):
This might need a bit more discussion around the date/time issue with Delius but I think this solves the problem for the appointments integration at least

src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
mahalrmoj: Will add:

    // nDelius stores all dates/times relative to Europe/London (without any timezone/offset information). Consumers
    // to community-api (e.g. interventions) can provide their dates/times with offsets from GMT (OffsetDateTime).
    // These methods carry out the necessary conversion to Europe/London times as required by delius-api.
    // The timezone used in the conversion is therefore set and must not be changed.

andymarke: @mahalrmoj I'm still trying to get me head around this:

So without this change if a consumer for instance called the endpoint with 10am 1/1/2020 (during GMT) and then sent 10 am 1/6/202 (during BST) was Delius then recording the second time as 9am? Is this what was being observed?

src/main/resources/application.yml
```diff
@@ -110,10 +110,10 @@ delius-integration-context:
       nsi-mapping:
         nsi-status: INPROG
         service-category-to-nsi-type:
-          "[Accommodation]": CRS01
-          "[Education, Training and Employment (ETE)]": CRS02
-          "[Finance, Benefits and Debt (FBD)]": CRS03
-          "[Dependency and Recovery (D&R)]": CRS04
+          428ee70f-3001-4399-95a6-ad25eaaede16: CRS01
```
andymarke, line 113: With this change does the endpoint now accept the key, so do clients send `428ee70f-3001-4399-95a6-ad25eaaede16` in the message? I'm struggling to understand why the key isn't just CRS03 or whatever?

andymarke (dismissed):
LGTM

src/main/resources/application.yml
mahalrmoj: Interventions has no knowledge of nsi types (e.g.CRS03) - it's a Delius specific concept. Interventions deals with service categories.

src/main/java/uk/gov/justice/digital/delius/utils/DateConverter.java
mahalrmoj: Yes that's right and specifically
1) Before BST and passes
```
        assertThat(OffsetDateTime.of(2021, 1, 1, 10, 0, 0, 0, ZoneOffset.UTC).toLocalDateTime())
            .isEqualTo(LocalDateTime.of(2021, 1, 1, 10, 0, 0, 0));
```
2) During BST 
```
        assertThat(OffsetDateTime.of(2021, 6, 1, 10, 0, 0, 0, ZoneOffset.UTC).toLocalDateTime())
            .isEqualTo(LocalDateTime.of(2021, 6, 1, 11, 0, 0, 0));
Expecting:
 <2021-06-01T10:00 (java.time.LocalDateTime)>
to be equal to:
 <2021-06-01T11:00 (java.time.LocalDateTime)>
```

src/main/resources/application.yml
andymarke: So still confused where the actual keys are referenced, are the sent in the body of the request?

mahalrmoj: Yes it is sent in the request from interventions

andymarke: Thanks, got it!

andymarke (approved):
LGTM

PR #406, opened 2021-04-20
🐛 PIC-1314 restore NSI subtype

chris-faulkner-mojdt:


mahalrmoj (approved):
Looks good. It was also taken out of NewNsi if you need that reinstated

PR #411, opened 2021-04-22
DAPI-170 When creating appointments introduced the distinction between a consumer requiring additional processing and one that doesn't

mahalrmoj:
Introduced the distinction between a client that requires additional processing and and one that doesn't. Each has its own endpoint for create appointment which shares same service implementation. Where additional processing is required, an integration context name must be supplied.

@RequestMapping(value = "/offenders/crn/{crn}/sentence/{sentenceId}/appointments" ...
@RequestMapping(value = "/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/{contextName}" ...

The additionally processing identified by contextName is in the form of
1) find a requirement for the sentence provided
2) use defaults as identified by the integration context

This enables community-api to support consumers with differing needs whilst sharing the same implementation.

**Is this a breaking change?**
Yes and limited to an endpoint used by interventions which is deemed work in progress.
"/offenders/crn/{crn}/sentence/{sentenceId}/appointments"

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -17,7 +17,7 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.justice.digital.delius.controller.advice.ErrorResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateResponse;
-import uk.gov.justice.digital.delius.data.api.AppointmentCreateWkcRequest;
+import uk.gov.justice.digital.delius.data.api.WellKnownAppointmentCreateRequest;
 import uk.gov.justice.digital.delius.service.AppointmentService;
 
 @RestController
@@ -65,9 +65,9 @@ public class AppointmentBookingController {
     @ApiOperation(value = "Creates an Contact appointment for a well known client")
     public ResponseEntity<AppointmentCreateResponse> createAppointmentWkc(final @PathVariable("crn") String crn,
                                                                           final @PathVariable("sentenceId") Long sentenceId,
-                                                                          final @RequestBody AppointmentCreateWkcRequest appointmentCreateWkcRequest) {
+                                                                          final @RequestBody WellKnownAppointmentCreateRequest wellKnownAppointmentCreateRequest) {
 
-        AppointmentCreateResponse response = appointmentService.createAppointment(crn, sentenceId, appointmentCreateWkcRequest);
+        AppointmentCreateResponse response = appointmentService.createAppointment(crn, sentenceId, wellKnownAppointmentCreateRequest);
         return new ResponseEntity<>(response, HttpStatus.CREATED);
     }
 }
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -17,6 +17,7 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.justice.digital.delius.controller.advice.ErrorResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateResponse;
+import uk.gov.justice.digital.delius.data.api.AppointmentCreateWkcRequest;
 import uk.gov.justice.digital.delius.service.AppointmentService;
 
 @RestController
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateWkcRequest.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotNull;
+import java.time.OffsetDateTime;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class AppointmentCreateWkcRequest {
```
axle-h, line 16: WKC: well known contact? can I suggest WellKnownAppointmentCreateRequest ?

src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateRequest.java
```diff
@@ -15,6 +15,14 @@
 @AllArgsConstructor
 public class AppointmentCreateRequest {
 
+    @NotNull
+    @ApiModelProperty(required = true)
```
axle-h, line 19: I dont think this should be required at this stage, pretty sure we discussed that not all appointments will be rar appointments?

electrified: Do we need the NSI Id field too?

Thinking we will link to the NSI OR the requirement (but never the event or the offender)

src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateWkcRequest.java
mahalrmoj: Will do

src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateRequest.java
mahalrmoj: OK I will relax it the constraint here.

WRT NSI that's not a requirement for interventions (yet) so I suggest if you need it you can add it as part of your evolution of the endpoint.

src/test/java/uk/gov/justice/digital/delius/service/AppointmentServiceTest.java
```diff
@@ -26,8 +27,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
electrified, line 30: IDE settings? I Think we normally go for no wildcard imports

electrified (approved):

src/test/java/uk/gov/justice/digital/delius/service/AppointmentServiceTest.java
mahalrmoj: will change

mxco86 (dismissed):
Look good - providing the generic endpoints and then wrapping with the more specific well-known ones makes sense to me. Would be good to get some more eyes on it before we merge.

src/test/java/uk/gov/justice/digital/delius/service/AppointmentServiceTest.java
```diff
@@ -80,6 +82,25 @@ public void appointmentsSortedByContactDateDescending() {
 
     @Test
     public void createsAppointment() {
+        // Given
+        var startTime = OffsetDateTime.now().truncatedTo(ChronoUnit.SECONDS);
+        var endTime = startTime.plusHours(1);
+
+        NewContact deliusNewContactRequest = aDeliusNewContactRequest(startTime, endTime);
```
andymarke, line 89: very minor, but prefer `final var`  rather than types to cut down on code noise

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
```diff
@@ -48,4 +49,25 @@
         AppointmentCreateResponse response = appointmentService.createAppointment(crn, sentenceId, appointmentCreateRequest);
         return new ResponseEntity<>(response, HttpStatus.CREATED);
     }
+
+    @RequestMapping(value = "/offenders/crn/{crn}/sentence/{sentenceId}/well-known/appointments",
+                method = RequestMethod.POST,
+                consumes = "application/json")
```
andymarke, line 55: Not 100% sure of this but I wonder if this would have been better modelled as

`/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/:context`

So move context out of the body and as a parameter?
rather than "well-known".

Since isn't essentially  "I want to create a appointment using this context" ?

andymarke (approved):
LGTM

src/test/java/uk/gov/justice/digital/delius/service/AppointmentServiceTest.java
mahalrmoj: Will do

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
mahalrmoj: We like that idea a lot. Will do

mxco86 (approved):
👍

andymarke (approved):
LGTM

PR #412, opened 2021-04-23
Update field description in RiskResourcingDetails

ollietreend:
The field description was originally annotated as a `note` – but that meant it wasn't visible in the Swagger UI docs.

This commit moves the description to the `value` property, which makes it visible.


andymarke (approved):
LGTM

andymarke:
Replaced with branch that had permissions to merge. Closing

PR #417, opened 2021-04-29
DAPI-171 Update of Appointment with Outcome for Context of CRS

mahalrmoj:
Introduced a new endpoint to carry out an appointment with an outcome type - for a specified context, e.g. commissioned rehabilitation services. Context specific PATCH hasn't been used after further technical and requirements discussions.

@RequestMapping(value = "/offenders/crn/{crn}/appointments/{appointmentId}/outcome/context/{contextName}",
        method = RequestMethod.POST, consumes = "application/json")

To make the update compatible with Delius the following transformation is done:
1) attended (Yes/No/Late) and notifyPPOfAttendanceBehaviour (true/false) are mapped to a single replace operation on outcome (e.g. AFTC)

This enables consumers of community-api to update Delius with the outcome of an appointment.

**Is this a breaking change?**
No.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -86,12 +86,11 @@ public class AppointmentBookingController {
         })
 
     @ApiOperation(value = "Updates an Contact appointment")
-    public ResponseEntity<AppointmentUpdateResponse> patchAppointmentWithContext(final @PathVariable("crn") String crn,
-                                                                                 final @PathVariable("appointmentId") Long appointmentId,
-                                                                                 final @PathVariable("contextName") String context,
-                                                                                 final @RequestBody JsonPatch jsonPatch) {
+    public AppointmentUpdateResponse patchAppointmentWithContext(final @PathVariable("crn") String crn,
+                                                                 final @PathVariable("appointmentId") Long appointmentId,
+                                                                 final @PathVariable("contextName") String context,
+                                                                 final @RequestBody JsonPatch jsonPatch) {
 
-        AppointmentUpdateResponse response = appointmentService.patchAppointment(crn, appointmentId, context, jsonPatch);
-        return new ResponseEntity<>(response, HttpStatus.OK);
+        return appointmentService.patchAppointment(crn, appointmentId, context, jsonPatch);
     }
 }
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.delius.controller.secure;
 
+import com.github.fge.jsonpatch.JsonPatch;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiResponse;
@@ -17,6 +18,7 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.justice.digital.delius.controller.advice.ErrorResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateResponse;
+import uk.gov.justice.digital.delius.data.api.AppointmentUpdateResponse;
 import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentCreateRequest;
 import uk.gov.justice.digital.delius.service.AppointmentService;
 
@@ -70,5 +72,26 @@ public class AppointmentBookingController {
 
         AppointmentCreateResponse response = appointmentService.createAppointment(crn, sentenceId, contextName, contextlessAppointmentCreateRequest);
         return new ResponseEntity<>(response, HttpStatus.CREATED);
+    }
+
+    @RequestMapping(value = "/offenders/crn/{crn}/appointments/{appointmentId}/context/{contextName}",
+        method = RequestMethod.PATCH,
+        consumes = "application/json")
+    @ApiResponses(
+        value = {
+            @ApiResponse(code = 200, message = "Updated", response = String.class),
+            @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+            @ApiResponse(code = 403, message = "Requires role ROLE_COMMUNITY_INTERVENTIONS_UPDATE"),
+            @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)
+        })
+
+    @ApiOperation(value = "Updates an Contact appointment")
+    public ResponseEntity<AppointmentUpdateResponse> patchAppointmentWithContext(final @PathVariable("crn") String crn,
+                                                                                 final @PathVariable("appointmentId") Long appointmentId,
+                                                                                 final @PathVariable("contextName") String context,
+                                                                                 final @RequestBody JsonPatch jsonPatch) {
+
+        AppointmentUpdateResponse response = appointmentService.patchAppointment(crn, appointmentId, context, jsonPatch);
+        return new ResponseEntity<>(response, HttpStatus.OK);
     }
 }
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -20,6 +20,7 @@ import uk.gov.justice.digital.delius.data.api.AppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentUpdateResponse;
 import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentCreateRequest;
+import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentOutcomeRequest;
 import uk.gov.justice.digital.delius.service.AppointmentService;
 
 @RestController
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
```diff
@@ -71,4 +73,25 @@
...
+        })
+
+    @ApiOperation(value = "Updates an Contact appointment")
+    public ResponseEntity<AppointmentUpdateResponse> patchAppointmentWithContext(final @PathVariable("crn") String crn,
+                                                                                 final @PathVariable("appointmentId") Long appointmentId,
+                                                                                 final @PathVariable("contextName") String context,
+                                                                                 final @RequestBody JsonPatch jsonPatch) {
+
+        AppointmentUpdateResponse response = appointmentService.patchAppointment(crn, appointmentId, context, jsonPatch);
+        return new ResponseEntity<>(response, HttpStatus.OK);
```
andymarke, line 95: You could yes make this function return `AppointmentUpdateResponse` - no need for the ResponseEntity

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
```diff
@@ -71,4 +73,25 @@
...
+    }
 }
```
andymarke, line 97: No sure the advantage to this compared to say a PUT with a DTO that has `notifyPPOfAttendanceBehaviour`, `notes`, `attended` or whatever. Feels this requires a greater cognitive load and harder to document.

But happy to see how this plays out

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
```diff
@@ -0,0 +1,100 @@
...
+
+import static com.fasterxml.jackson.databind.node.TextNode.valueOf;
+import static com.github.fge.jackson.jsonpointer.JsonPointer.of;
+import static java.lang.String.format;
+import static java.util.Optional.ofNullable;
+import static java.util.stream.Collectors.toList;
+
+@Component
+@AllArgsConstructor
+public class AppointmentPatchRequestTransformer {
```
andymarke, line 25: See note above - but feels like all this code is required since we went down the PATCH route rather than PUT.

Again happy to see how this plays out, but this amount a code suggests it is overkill

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
mahalrmoj: Will do

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
mxco86: Yes introducing PATCH methods is something we discussed a lot with the probation TAs and devs. We're using this pattern where the client doesn't have enough information to construct a full PUT. It's probably worth a proper discussion on the approach and reasoning if we can make some time.

src/main/java/uk/gov/justice/digital/delius/service/AppointmentService.java
```diff
@@ -58,21 +48,34 @@ public AppointmentService(ContactTypeRepository contactTypeRepository,
...
+        final var request = appointmentOf(contextualRequest, requirement.getRequirementId(), context);
 
         return createAppointment(crn, sentenceId, request);
     }
 
+    public AppointmentUpdateResponse patchAppointment(String crn, Long appointmentId, JsonPatch jsonPatch) {
+
+        final var contactDto = deliusApiClient.patchContact(appointmentId, jsonPatch);
+        return new AppointmentUpdateResponse(contactDto.getId());
+    }
```
axle-h, line 70: I think it would make sense here to assert that the contact is an appointment type. Otherwise this endpoint could be abused.

axle-h, line 70: are we cleanly catching 400 errors from the delius api? ie I would expect them to be passed through.

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private static final String SOURCE_ATTENDED_FIELD_PATH = "/attended";
+    private static final String SOURCE_NOTIFY_PP_FIELD_PATH = "/notifyPPOfAttendanceBehaviour";
+    private static final String TARGET_OUTCOME_FIELD_NAME = "outcome";
+
+    private final ObjectMapper objectMapper;
+
+    public JsonPatch mapAttendanceFieldsToOutcomeOf(final JsonPatch jsonPatch, final IntegrationContext context) {
+
+        var nodes = objectMapper.convertValue(jsonPatch, JsonNode.class);
```
axle-h, line 38: see you're converting to json here & using the jackson api to mutate the patch document. Is this required? can't you just copy & mutate the JsonPatch structure in place? I feel a lot of the code here is spent on navigating the serialised json patch structure for which we have this JsonPatch abstraction.

src/main/java/uk/gov/justice/digital/delius/service/AppointmentService.java
mahalrmoj: Yes I has tested that they are passed through successfully.

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
axle-h: ergh I just had a look at the API for JsonPatch and I'm not sure this is possible. What a shame.

src/main/java/uk/gov/justice/digital/delius/service/AppointmentService.java
mahalrmoj: Do this require two checks:
1. The underlying contact has the correct appointment type
2. The patch is not changing the appointment type

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
mahalrmoj: Found very little support for modifying a JsonPatch in the way we wanted. Will talk to you about what we can do instead.

mxco86 (dismissed):
This looks fine to me as an initial implementation. Hopefully we'll be able to simplify it a bit as we work on it over time, especially the transforming code.

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
mahalrmoj: Covered by previous comment

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
```diff
@@ -0,0 +1,100 @@
...
+                                                                             final List<JsonPatchOperation> originalReplaceOperations) {
+
+        var replaceOperations = new ArrayList<>(originalReplaceOperations);
+
+        if ( attendedNode.isPresent() ) {
+            var mappings = context.getContactMapping()
+                .getAttendanceAndBehaviourNotifiedMappingToOutcomeType();
+
+            var attendedType = attendedNode.orElseThrow(
+                () -> new IllegalStateException("Attended does not exist in json patch"));
```
mikehalmamoj, line 71: I don't really understand the reasoning behind choosing `PATCH` so forgive me if my viewpoint is naive... but don't we lose any possibility of documenting business rules like this in the OpenApi documentation by choosing PATCH?

My gut feeling is that it would be better to use a `PUT` or `POST` with a request object that contains the information the client knows about, then we can put the business rules in the API and lose quite a lot of complicated code.  I guess you can work out what should go in the `JsonPatch` by reading the tests but... :man_shrugging:

mxco86: Would your concerns be less if there was a GET endpoint the retrieve the appointment resource? This would give the resource representation that is being patched and possibly make things a lot clearer. The fact that we don't have this means we're in a slightly strange state with the resource modelling of appointments. I take the point that the transformation code is not optimal and we will need to work to improve this.

axle-h (dismissed):

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
mahalrmoj: Now using POST operation with DTO

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
mahalrmoj: Complexity now removed as source is a DTO

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
mahalrmoj: Moved away from patch as made the endpoint a far more business specific operation on an appointment. Agree, much more self documenting and simpler.

mxco86 (approved):
👍

axle-h (approved):

src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessAppointmentOutcomeRequest.java
```diff
@@ -0,0 +1,29 @@
...
+    private String notes;
+
+    @NotNull
+    @ApiModelProperty(required = true)
+    private String attended;
+
+    @NotNull
+    @ApiModelProperty(required = true)
+    private Boolean notifyPPOfAttendanceBehaviour;
+}
```
andymarke, line 29: curious, what is the `PP` in `notifyPPOfAttendanceBehaviour`?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessAppointmentOutcomeRequest.java
mahalrmoj: Probation Practitioner

PR #418, opened 2021-04-30
E2E-347 - Add additional fields to appointment creation response

electrified:


src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentCreateResponse.java
```diff
@@ -15,4 +16,8 @@
 
     @NotNull
     private Long appointmentId;
+    private OffsetDateTime appointmentStart;
```
mahalrmoj, line 19: Should these be marked as @NotNull and if not should we remove the @NotNull from appointmentId?

electrified: Thanks, I will add it in for all of them.

endTime is optional on the DAPI side (only needed if an outcome is present I believe), but as we are always supplying one, we should always get one back

axle-h (approved):

andymarke (approved):
LGTM

PR #422, opened 2021-05-07
DAPI-171 Correction to configuration

mahalrmoj:
Correction to configuration for mapping attended/behaviour data to outcomes. Yes/No get translated to true and false by spring. Need to be quoted.

Correction.

**Breaking Changes?**
None


mxco86 (approved):
Could you add a comment that YAML requires the quoting on "yes" and "no" so it's clear that these are necessary

mxco86 (approved):

PR #423, opened 2021-05-10
DAPI-172 DAPI-173 Changes to Referral Start endpoint for interventions

mahalrmoj:
After a review of the interaction with Interventions, a number of changes where identified for the referral sent end point:

1) A more generic name for the endpoint is ReferralStart
2) Referrals are not limited to Sentences with RAR requirements, and thus, exception not raised if requirement not found
2) ContractType is to be used to map to NsiType instead of ServiceCategory
3) SentAt is renamed to StartedAt
4) Context name is to be specified in URL as apposed to in the body - in line with appointment creation

This change means that the initial release of interventions can be supported.

**Breaking Changes?**
Yes and limited to interventions referralSent endpoint.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessReferralStartRequest.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessReferralStartRequest.java
@@ -17,15 +17,15 @@ import java.util.UUID;
 @Builder(toBuilder = true)
 @NoArgsConstructor
 @AllArgsConstructor
-public class ReferralSentRequest {
+public class ContextlessReferralStartRequest {
 
     @NotNull
     @ApiModelProperty(required = true)
-    private OffsetDateTime sentAt;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessReferralStartRequest.java
```diff
@@ -17,15 +17,15 @@
...
 
     @NotNull
     @ApiModelProperty(required = true)
-    private OffsetDateTime sentAt;
+    private OffsetDateTime startedAt;
 
     @NotNull
     @ApiModelProperty(required = true)
-    private UUID serviceCategoryId;
+    private String contractType;
```
electrified, line 28: I wonder if this field could also do with documenting as it's not part of the usual NDelius vocabulary?

electrified (approved):

electrified (approved):

PR #425, opened 2021-05-10
MAS-415 Error calculating tier when requirements request fails

nickmcmahon01:
* Lazy Fetch OGRSAssessemnt
* move the ActiveOnly filtering for events into the Sql not codespace


src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -38,27 +38,17 @@
 
     public ConvictionRequirements getRequirementsByConvictionId(String crn, Long convictionId, boolean activeOnly) {
 
-        Predicate<Requirement> activeFilter = activeFilter(activeOnly);
-        var requirements = Optional.of(getEvent(crn, convictionId))
+        var requirements = Optional.of(getEvent(crn, convictionId, activeOnly))
```
markreesmoj, line 41: 👎 
The activeOnly filter must be applied to requirements. We can try to additionally apply it to events and disposals to see if it helps get round this problem but it must be on requirements - that's what we are relying on the method to do. Otherwise we (and others?) will start getting inactive requirements again

markreesmoj (changes requested):

PR #427, opened 2021-05-11
Dapi 174 175 create appt with contract type rar and nsi lookup

mahalrmoj:


axle-h:
shouldn't 'contract' type be 'contact' type? Or is that something else?

mahalrmoj:
> shouldn't 'contract' type be 'contact' type? Or is that something else?

WRT interventions (CRS), we must associate the new appointment to an existing NSI. However, Interventions doesn't hold the NSI Id. To find the existing NSI Id in community-api, we need to search based on criteria, one of which is NSI Type. To determine NSI Type, we map from Contract Type. A a bit protracted.

andymarke (dismissed):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/ContextlessAppointmentCreateRequest.java
```diff
@@ -17,17 +17,28 @@
 
     @NotNull
     @ApiModelProperty(required = true)
-    private OffsetDateTime appointmentStart;
+    private String contractType;
 
     @NotNull
     @ApiModelProperty(required = true)
-    private OffsetDateTime appointmentEnd;
+    private OffsetDateTime referralStart;
```
mxco86, line 24: Is it worth grouping the context and detail of the contextless requests using a couple of comments?

mahalrmoj: Sounds sensible to me - will do

mxco86 (approved):
👍

axle-h (approved):

PR #431, opened 2021-05-11
MAS-415 only retrieve active events when getting active requirements

markreesmoj:
We have a live incident where our request for requirements is failing because of a Disposal with two unpaid work records. This is incorrect historic data which hopefully can not happen on new Disposals, but to avoid it happening again, we want to avoid retrieving inactive Events when retrieving active requirements


src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -78,16 +81,22 @@ public PssRequirements getPssRequirementsByConvictionId(String crn, Long convict
...
+        return eventRepository.findByOffenderId(offenderId)
                 .stream()
                 .filter(event -> convictionId.equals(event.getEventId()))
                 .findAny()
                 .orElseThrow(() ->  new NotFoundException(format("Conviction with convictionId '%s' not found", convictionId)));
     }
 
-    private Offender getOffender(String crn) {
-        return offenderRepository.findByCrn(crn)
+    private Event getActiveEvent(String crn, Long convictionId) {
```
markreesmoj, line 92: This used to be like the method above, which seems a wasteful way to get an Event which we have the ID of

src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -78,16 +81,22 @@ public PssRequirements getPssRequirementsByConvictionId(String crn, Long convict
...
+        var offenderId = getOffenderId(crn);
+        return eventRepository.findByOffenderIdAndEventIdAndActiveTrue(offenderId, convictionId)
+            .orElseThrow(() ->  new NotFoundException(format("Active conviction with convictionId '%s' not found", convictionId)));
+    }
+
+    private Long getOffenderId(String crn) {
```
markreesmoj, line 98: We only ever want the offenderID in this service

andymarke (approved):
LGTM

PR #433, opened 2021-05-11
Dt 1780 insert court poc

andymarke:
This is initial skeleton of  a proof of concept for a centralised court register held outside of Delius and NOMIS.

These skeletons will update/insert courts but if the POC goes further the actually updates will be moved to the delius-api.

Just skeleton endpoints for now, but what is of technical interest is one use a DTO as a java `record` the other uses lombok.

My view is lombok is better than java records due to the lack of a `copy` function and no `builder` methods so you sendup with a nasty huge constructors without the ability to name parameters.


build.gradle
```diff
@@ -33,6 +33,11 @@ dependencies {
     implementation 'io.vavr:vavr:0.10.3'
     implementation 'com.github.java-json-tools:json-patch:1.13'
 
+    // needed for record serialisation
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.12.3'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.3'
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.12.3'
+
```
andymarke, line 40: upgrade required for java record serialisation

src/main/java/uk/gov/justice/digital/delius/data/api/NewCourtDto.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.justice.digital.delius.data.api;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import javax.validation.constraints.NotBlank;
+
```
andymarke, line 8: Note the use of `record` - see main comment where I don't see enough benefits to use this -  the 2 big disadvantages you will see in the unit tests

1. No ability to copy and existing instance and copy (e.g like kotlin `copy` or lombok `toBuilder`
2. No ability to name constructor parameters - so you end up with a huge constructor where parameter ordering matters (unlike kotlin named parameters or a builder pattern)

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/CourtAPITest.java
```diff
@@ -0,0 +1,284 @@
...
+                .statusCode(200);
+        }
+
+        private String validInsertRequest() {
+            return insertRequest("New Magistrates Court", NEW_COURT_CODE, "MAG", "N53");
+        }
+        private String insertRequest(String courtName, String code, String courtTypeCode, String probationArea) {
+            return writeValueAsString(new NewCourtDto(code, courtTypeCode, true, courtName, null, null, "Crown Square", "High Street", "Town Centre", "Sheffield", "South Yorkshire", "S1 2BJ", "England", probationArea));
+        }
+    }
```
andymarke, line 238: huge constructor - not very good IMHO

PR #444, opened 2021-05-17
DT-1780 Implementation of insert court details with feature switch

andymarke:
Experimental API to create Courts as part of the HMPPS REGISTERS project.

1. API is likely to be replaced with a delius-api implementation is the proof of concept is converted to a production ready project
2. Switched of in all environments except test where only creation court codes and be used


src/main/resources/db/schema/V1_0__schema.sql
```diff
@@ -388,7 +388,7 @@ create sequence COURT_APPEARANCE_ID_SEQ
     ;
 
 create sequence COURT_ID_SEQ
-    minvalue 2500000000
+    minvalue 2600000000
     maxvalue 900000000000000000
```
andymarke, line 392: value was conflicting with existing records in H2

lbennett-moj (approved):

PR #445, opened 2021-05-17
DT-1780 Get all courts and improve update errors

andymarke:
1. Use `Either` pattern rather then exception
2. Endpoint to get all courts


src/main/java/uk/gov/justice/digital/delius/service/CourtService.java
```diff
@@ -105,6 +106,19 @@ public Court getCourt(String code) {
         });
     }
 
+    public List<Court> getCourts() {
+        return courtRepository
+            .findAll()
+            .stream()
+            // filter out duplicates (see note above)
+            .filter(court -> court.getCourtId().equals(getMostLikelyCourt(court.getCode()).orElseThrow().getCourtId()))
```
mikehalmamoj, line 114: Do you need a `distinct()` after this?  It appears to replace courts with the most likely court, in which case they would be duplicated?

mikehalmamoj: Maybe not  - there's a test for it :)

mikehalmamoj (approved):

PR #446, opened 2021-05-18
DT-2032: 🐛 Change order of queries to avoid duplicates in oracle

petergphillips:


src/main/java/uk/gov/justice/digital/delius/service/UserService.java
```diff
@@ -66,15 +64,11 @@ public AccessLimitation accessLimitationOf(final String subject, final OffenderD
         return accessLimitationBuilder.build();
     }
 
-    private List<String> probationAreaCodesOf(final List<ProbationArea> probationAreas) {
```
petergphillips, line 69: private method not used

src/main/java/uk/gov/justice/digital/delius/service/UserService.java
```diff
@@ -66,15 +64,11 @@ public AccessLimitation accessLimitationOf(final String subject, final OffenderD
...
-                pas -> pas.stream().map(ProbationArea::getCode).collect(toList())).orElse(Collections.emptyList());
-    }
-
     public Optional<UserDetails> getUserDetails(final String username) {
-        final var oracleUser = userRepositoryWrapper.getUser(username);
-        return ldapRepository.getDeliusUser(username).map(user ->
-                UserDetails
+        final var ldapUser = ldapRepository.getDeliusUser(username);
+        return ldapUser.map(user -> {
+            final var oracleUser = userRepositoryWrapper.getUser(username);
```
petergphillips, line 70: switch to only checking oracle if user found in ldap

mikehalmamoj (approved):

PR #449, opened 2021-05-18
DT-2025: 🐛 Handle duplicates for inactive events in probation status

petergphillips:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -392,11 +392,10 @@ private ProbationStatus probationStatusOf(Offender offender) {
     }
 
     private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
-        if (!probationStatusOf(offender).equals(ProbationStatus.PREVIOUSLY_KNOWN)) {
+        if (probationStatusOf(offender) != ProbationStatus.PREVIOUSLY_KNOWN) {
```
petergphillips, line 395: this is an enum so no point doing `equals`

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -392,11 +392,10 @@ private ProbationStatus probationStatusOf(Offender offender) {
...
             return null;
         }
-        return Optional.ofNullable(offender.getEvents())
-            .orElse(Collections.emptyList())
+        return offender.getEvents()
```
petergphillips, line 398: pointless `ofNullable` since spring jpa guaranteed never to return `null` for collections

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -405,23 +404,19 @@ private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
     }
 
     private Boolean inBreachOf(Offender offender) {
-        if (!probationStatusOf(offender).equals(ProbationStatus.CURRENT)) {
+        if (probationStatusOf(offender) != ProbationStatus.CURRENT) {
```
petergphillips, line 407: again this is an enum so no point doing `equals`

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -405,23 +404,19 @@ private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
...
             return null;
         }
-        return Optional.of(offender)
```
petergphillips, line 411: can't see any reason to wrap this in an `Optional` since if `offender` is `null` `Optional.of` will throw a null pointer anyway

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -405,23 +404,19 @@ private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
...
-            .map(o ->
-                activeEvents(o.getEvents())
-                    .stream()
-                    .anyMatch(event -> event.getInBreach() == 1L)
-            ).orElse(null);
+        return offender.getActiveEvents()
```
petergphillips, line 410: change to use new `activeEvents` method

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -405,23 +404,19 @@ private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
...
+            .stream()
+            .anyMatch(event -> event.getInBreach() == 1L);
     }
 
     private Boolean preSentenceActivityOf(Offender offender) {
-        return Optional.of(offender)
```
petergphillips, line 420: again can't see any reason to wrap this in an `Optional` since if `offender` is `null` `Optional.of` will throw a null pointer anyway

src/main/resources/logback-spring.xml
```diff
@@ -37,17 +37,21 @@
...
-            <appender-ref ref="consoleAppender"/>
-        </logger>
+      <logger name="org.flywaydb" additivity="false" level="INFO">
+        <appender-ref ref="consoleAppender"/>
+      </logger>
 
-        <root level="INFO">
-            <appender-ref ref="consoleAppender"/>
-        </root>
+      <logger name="org.flywaydb.core.internal.sqlscript.DefaultSqlScriptExecutor" additivity="false" level="WARN">
```
petergphillips, line 48: this stops the count of rows affected being output, which for community api happens a lot as loads of data is inserted

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -405,23 +404,19 @@ private LocalDate previouslyKnownTerminationDateOf(Offender offender) {
...
-            .map((o) -> o.getEvents()
-                .stream()
-                .anyMatch(event -> event.getDisposal() == null)
-            ).orElse(false);
+        return offender.getActiveEvents()
```
petergphillips, line 416: change in behaviour here - only look at active events to see if a disposal is missing

src/main/resources/logback-spring.xml
```diff
@@ -45,6 +45,10 @@
             <appender-ref ref="consoleAppender"/>
         </logger>
 
+        <logger name="org.flywaydb.core.internal.sqlscript.DefaultSqlScriptExecutor" additivity="false" level="WARN">
```
petergphillips, line 48: these were annoying, change recently to flyway means that every sql script posts a count of rows affected, which for community api is lots

src/test/java/uk/gov/justice/digital/delius/service/ConvictionService_GetProbationStatusTest.java
```diff
@@ -165,13 +151,14 @@ public void canGetProbationStatusForPreviouslyKnownOffenderWithPreSentenceActivi
         when(offenderRepository.findByCrn(CRN)).thenReturn(Optional.of(offender));
         // PREVIOUSLY_KNOWN if currentDisposal is false and they have at least 1 event with a disposal
         when(offender.getCurrentDisposal()).thenReturn(0L);
+        when(offender.getActiveEvents()).thenReturn(List.of(event, event2));
         when(offender.getEvents()).thenReturn(List.of(event, event2));
```
mikehalmamoj, line 155: Did you forget to drop this line?

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
andymarke: @chris-faulkner-mojdt I think Peter is correct here, right now if there was offender who back in that day had a pre-sentence event raised which was then terminated without a sentence (e.g. they changed their plea and was found not guilty) the old code we treat that has "pre sentence" which is not correct since it is historic and there is no longer any pre-sentence active.

andymarke (approved):
LGTM

src/test/java/uk/gov/justice/digital/delius/service/ConvictionService_GetProbationStatusTest.java
petergphillips: Unfortunately not.  To work out the probation status and previously known termination date then we need to look at all the events, so therefore the code does indeed call both.

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
chris-faulkner-mojdt: Sure person understood this but there may be other reasons for choosing that method

petergphillips: Interesting.  I don't normally work in this code base, so are you suggesting that doing a `!` and `equals` for enums is preferable here?

PR #450, opened 2021-05-18
Dt 1780 testing issues

andymarke:
* Changed reference data lookup errors to 400 error (was 500) as convention 
* Changed url to plural `courts` from `court` as convention

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/CourtService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/CourtService.java
@@ -27,28 +27,35 @@ public class CourtService {
 
     @Transactional
     public Either<BadData, Court> updateCourt(String code, UpdateCourtDto court) {
-        return getMostLikelyCourt(code).map(courtEntity -> {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/digital/delius/service/CourtService.java
```diff
@@ -116,9 +134,41 @@ public Court getCourt(String code) {
...
+        }
     }
 
-    public static record CourtDoesNotExist(String courtCode) {
+    public static record ProbationDoesNotExist(String probationAreaCode) implements BadData {
+        public String message() {
+            return String.format("Probation area %s does not exist", probationAreaCode);
+        }
+    }
+
```
andymarke, line 167: small helper to give clarity to the mean of left and right when using Either library

mikehalmamoj (approved):

lbennett-moj (approved):

src/main/java/uk/gov/justice/digital/delius/service/CourtService.java
lbennett-moj: more readable

PR #454, opened 2021-05-20
DAPI-177 ReferralStart always creates a new NSI

mahalrmoj:
A new referral started in Interventions should always create a new NSI in Delius and not reuse an existing one should one exist. Also, to ensure the correct NSI is a matched with referral actions, status date time is also used to find the correct match.

Added additional Contract Type Procurement Codes mappings to NSI type in config.

There is a bug in the referral start endpoint which may associate an existing NSI with a new referral in Interventions.

Breaking Changes?
None


mahalrmoj:
Not safe to use StatusDateTime

PR #456, opened 2021-05-20
DT-2025: Switch event to use booleans

petergphillips:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Event.java
```diff
@@ -112,18 +123,18 @@
     private LocalDate cpsDate;
 
     @Column(name = "CPS_SOFT_DELETED")
-    private Long cpsSoftDeleted;
+    private Boolean cpsSoftDeleted;
```
petergphillips, line 126: note this changed to a `Boolean` since database is nullable

PR #458, opened 2021-05-21
DAPI-164 Mapping additional Contract Types to Delius Nsi Types via config

mahalrmoj:
Configuration to cover additional Contract Type or NSI Type mappings.

Support for additional Contract Types in Delius.

**Breaking Changes**
None


mxco86 (approved):
LGTM

PR #460, opened 2021-05-25
E2E-520: Get offender appointments

axle-h:
New API for getting offender appointments by CRN. Uses same URL structure as existing, /api endpoint.

image


andymarke (dismissed):
LGTM

axle-h:
@andymarke thanks for the review. I just tried consuming this via OAPI & realised I'd left the optional query parameters in as required... it's even in the screenshot...
So can I get a re-review plz.

PR #463, opened 2021-05-26
DAPI-177 When searching for a matching NSI using existing criteria, if multiple are found, then use ReferralId (URN) to find a unique one. 

mahalrmoj:
There is an edge case when more that one Referral may be created in R&M that appear to cover the same Intervention, i.e. are for the same CRN and Sentence, cover the same Contract Type (e.g. Personal Wellbeing) and start on the same date. Prior to this change they would both be tied to the same NSI in Delius. After this change, each one results in its own NSI using the ReferralId (in the form of an URN) to distinguish them ... should an initial search return multiple. The belief is this case is rare and so the additional filtering by URN will not come into play most of the time. There is no designated field for the URN to reside in so it is held in the Notes as an initial entry. The Notes field is never replaced, only appended too.

Specifically the following changes:
1. The Interventions ReferralId is stored as a URN against each NSI created
2. The URN is used as a last resort when finding the corresponding NSI in Delius, when multiple NIS's exist for the same CRN and Sentence, Contract Type and on the same Referral Sent Date.

To ensure that should multiple referrals be created which appear to be the same, they can be accounted for correctly by the calculations carried out in Delius.

**Breaking Changes?**
None

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
@@ -18,15 +18,21 @@ import uk.gov.justice.digital.delius.data.api.deliusapi.NewNsi;
 import uk.gov.justice.digital.delius.data.api.deliusapi.NewNsiManager;
 import uk.gov.justice.digital.delius.jpa.standard.repository.ContactRepository;
 import uk.gov.justice.digital.delius.transformers.NsiPatchRequestTransformer;
+import uk.gov.justice.digital.delius.transformers.ReferralTransformer;
 
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
+import javax.validation.constraints.NotNull;
 
 import static java.lang.String.format;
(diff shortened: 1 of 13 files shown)


andymarke (dismissed):
LGTM

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
```diff
@@ -126,11 +133,44 @@ public ReferralEndResponse endNsiReferral(final String crn, final String context
...
+            existingNsis = findExistingNsiByReferralUrnIfSupplied(existingNsis, referralId);
         }
         return existingNsis.stream().findFirst();
     }
 
+    /**
+     * There is an edge case when more that one Referral may be created in R&M that appear to cover the same Intervention,
+     * i.e. are for the same CRN and Sentence, cover the same Contract Type (e.g. Personal Wellbeing) and start on the same date.
+     * It is suspected that this is a rare scenario but can arise.
+     * This method comes into play when that situation arises and uses the referral id (in the form f a URN) to carry out a
```
mxco86, line 145: 👍 for the explanation of the edge case. Tiny typo - 'in the form of'

mxco86 (approved):
LGTM

mxco86 (approved):
LGTM

PR #466, opened 2021-05-28
✨ PIC-1426 Add "awaitingPsr" to ProbationStatusDetail and populate

chris-faulkner-mojdt:


src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
```diff
@@ -419,6 +421,17 @@ private Boolean preSentenceActivityOf(Offender offender) {
...
+    private Boolean awaitingPsrOf(Offender offender) {
+
+        return offender.getActiveEvents()
+            .stream()
+            .filter(event -> event.getDisposal() == null)
+            .map(Event::getCourtAppearances)
+            .flatMap(Collection::stream)
+            .map(CourtAppearance::getOutcome)
+            .anyMatch(outcome -> ReferenceDataService.REFERENCE_DATA_PSR_ADJOURNED_CODE.equals(outcome.getCodeValue()));
+    }
```
andymarke, line 433: Have you checked this in Delius?
Just thinking about this simple scenario 
1) Case adjourned for PSR
2) PSR delivered
3) Last court appearance then has a sentence outcome 

In that simple most common scenario wouldn't the above incorrectly return true?

Don't you need the last court appearance for an event not any?

andymarke (approved):
Not sure this logic is correct?

src/main/java/uk/gov/justice/digital/delius/service/ConvictionService.java
chris-faulkner-mojdt: Wouldn't that event then have a sentence though ? Agreed on the latest court appearance. 
Should it be - latest court appearance, no sentence, outcome "REFERENCE_DATA_PSR_ADJOURNED_CODE: ?

jhevans: We had a bit of back and forth about this when refining, initially I specced it up as you describe @andymarke but we changed it to the logic Chris has implemented because it didn't quite meet the user need. 

The rationale is that what the user is interested in regarding PSR's is whether they need to check if a PSR is outstanding or was requested so they can check if it's been done and get it done if not. In the majority of cases this will indeed mean that the last court appearance was 'Adjourned for PSR' - but if the case is then adjourned again for some other reason we still want to show the PSR badge in the front end in case the PSR still hasn't been done for whatever reason.

andymarke: Yes @chris-faulkner-mojdt you are right. And agree @jhevans the risk is PSR has been delivered but there are more adjournments  for other other reasons.

It would be really interesting to see in real world how often the badge really represents what is the real state is (as opposed to the state in the Delius database)

PR #467, opened 2021-06-01
E2E-521: Get offender contact summaries by CRN

axle-h:
New API for retrieving a subset of contact data by offender CRN.

image


src/main/java/uk/gov/justice/digital/delius/data/api/ContactSummary.java
```diff
@@ -0,0 +1,54 @@
...
+public class ContactSummary {
+    @NotNull
+    @ApiModelProperty(name = "Contact id", required = true)
+    private Long contactId;
+
+    @ApiModelProperty(name = "Contact start date & time", example = "2021-05-25T10:00:00+01:00")
+    private OffsetDateTime contactStart;
+
+    @ApiModelProperty(name = "Contact end date & time", example = "2021-05-25T11:00:00+01:00")
+    private OffsetDateTime contactEnd;
```
electrified, line 25: With Rajiv's work we continued the pattern of Community API only using LocalDateTimes on it's exposed APIs, using offsets would be a departure from that (although I agree offsets are better).

src/main/java/uk/gov/justice/digital/delius/transformers/ContactTransformer.java
```diff
@@ -178,4 +202,11 @@ private static KeyValue contactOutcomeTypeOf(ContactOutcomeType contactOutcomeTy
                         .build()).orElse(null);
     }
 
+    private static OffsetDateTime offsetDateTimeOf(LocalDate date, LocalTime time) {
```
electrified, line 205: There is a DateConverter class with other similar methods - could put it there?

src/main/java/uk/gov/justice/digital/delius/data/api/ContactSummary.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotNull;
+import java.time.OffsetDateTime;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class ContactSummary {
```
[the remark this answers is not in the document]
axle-h: It's not strictly necessary, it's an efficiency & standards thing. Contact has some really big stuff on there that we do not need e.g. NSI, license condition. It more inline datatype wise with the appointment model. The Contact model is basically a raw contact.

src/main/java/uk/gov/justice/digital/delius/data/api/ContactSummary.java
axle-h: There are still offset date times all over the appointment API, which is why they were used here. Is there work outstanding to convert to locals?

electrified: OK - Ignore me. I though we were using offsets on DAPI and locals on CAPI but it's the other way around!

andymarke: Is it possible to explain what the issue is with using LocalDateTime? What would be scenario we would get the wrong time given we always run in Europe/London timezone? Are contact times wrong by an hour in summer or something?

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderContactSummariesByCrn.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.List;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.RestAssured.withArgs;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.notNullValue;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
andymarke, line 18: very minor, you could omit SpringBootTest it is one the base class

andymarke (dismissed):
LGTM

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_getOffenderContactSummariesByCrn.java
axle-h: TBH, Im not actually sure what the issue with LocalDateTime is - I suppose the UTC -> GMT would have to be done in the client if we didn't do it here. At this point I've only used it to remain consistent with the existing appointment APIs created by interventions team. What are your thoughts on it?

src/main/java/uk/gov/justice/digital/delius/data/api/ContactSummary.java
mxco86: For appointments we were trying to use an unambiguous time value as far as we could as the data crosses a number of system boundaries. The source system (Interventions) is using UTC and the ultimate destination (Delius) is un-timezoned (!). Relying on the environment timezone of the three intermediate systems seems brittle, especially as they are deployed in different environments, however, I appreciate that this is not something that can be fixed without coordination across teams and using OffsetDateTime is inconsistent with existing CAPI code. Maybe this is something that needs a wider conversation?

andymarke: I remember for Alfresco someone had to use OffsetDateTime which I guess suggests Alfresco was setup differently to Delius.

TBH after all these years timezones *still* confuse me. Which is why I like to see a "test case" to understand the real world issue. And it sounds like this is an issue that *might* happen if Delius was change timezones.

mxco86: Yes timezones are the source of ultimate confusion for me too :) This doesn't fix the overall problem but it does take out one external variable which we saw as a positive. The test case would be an end-to-end test across systems where the timezones of each system env are the same, one is different, two are different etc.

andymarke (approved):
LGTM

PR #468, opened 2021-06-02
E2E-489 Add disability adjustments

electrified:


axle-h (approved):

andymarke:
Can you see what has happened to RegistrationRepostory ? GitHub is saying the changes are too many to display, IntelliJ thinks there are changes but can't find any 🤷🏽‍♂️

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Disability.java
```diff
@@ -28,6 +29,8 @@
     private LocalDate finishDate;
     @Column(name = "NOTES")
     private String notes;
+    @OneToMany(mappedBy = "disability", fetch = FetchType.LAZY)
+    private List<Provision> provisions;
```
andymarke, line 33: We had an interesting issue where it looked like the LAZY was not being honoured because it was a `mapped` relationship

andymarke (approved):
LGTM

@electrified can we schedule in a small bit of analysis after this goes live to see if there is any performance impact (which I suspect there isn't).

What I am thinking this is a very specific bit of data that I suspect very few clients are interested in so there could be an argument to add a new endpoint.

However that in itself means at least 2 calls to get that is data.

So it would be interesting to see if there is any noticeable differences with performance with this extra call. If there isn't it would guide us with future changes like this

electrified:
> Can you see what has happened to RegistrationRepostory ? GitHub is saying the changes are too many to display, IntelliJ thinks there are changes but can't find any 🤷🏽‍♂️

I've reverted the changes to RegistrationRepostory.

src/main/java/uk/gov/justice/digital/delius/data/api/Disability.java
```diff
@@ -17,4 +18,5 @@
     private LocalDate startDate;
     private LocalDate endDate;
     private String notes;
+    private List<Provision> provisions;
```
andymarke, line 21: can you think of a better name for this? I can guess this means provisions to mitigate the disability but not 100% sure. Don't worry if you can't think of a better name.

As a matter of interest do you have an example for a provision?

electrified: On the NDelius frontend, these are called Adjustments rather than provisions - I've gone with naming the same as the database table, as I think there is precedent for that (e.g. Probation Area is Provider on the front end, but we still call it Probation Area here)

Examples could be something physical like a wheelchair ramp or stair lift, or additional help/ assistance.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Disability.java
electrified: I'll have a look at the queries being generated.

PR #473, opened 2021-06-07
Update 'user access' endpoints to be more flexible 

tomsmyers:
this change has two parts

1. change the permissions on the existing `/offenders/crn/{crn}/userAccess` endpoint to allow it to be called with delius user tokens (ROLE_PROBATION).
2. add a new endpoint which allows specifying the username in the url to allow regular client credentials granted tokens the ability to use the endpoint without requesting a new 'username' scoped token.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -614,6 +614,7 @@ public class OffendersResource {
             @ApiResponse(code = 403, message = "User is restricted from access to offender", response = AccessLimitation.class),
             @ApiResponse(code = 404, message = "No such offender, or no such User (see body for detail)")
     })
+    @PreAuthorize("hasRole('ROLE_COMMUNITY') || hasRole('ROLE_PROBATION')")
     public ResponseEntity<AccessLimitation> checkUserAccessByCrn(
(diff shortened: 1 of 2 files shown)


andymarke (dismissed):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -614,14 +614,27 @@ public SentenceStatus getSentenceStatusBySentenceId(
             @ApiResponse(code = 403, message = "User is restricted from access to offender", response = AccessLimitation.class),
             @ApiResponse(code = 404, message = "No such offender, or no such User (see body for detail)")
     })
+    @PreAuthorize("hasRole('ROLE_COMMUNITY') || hasRole('ROLE_PROBATION')")
```
petergphillips, line 617: can do `hasAnyRole` such as https://github.com/ministryofjustice/hmpps-auth/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/resource/api/AuthUserRolesController.kt#L84

tomsmyers: nice - thanks. updated.

andymarke (approved):
LGTM

PR #476, opened 2021-06-09
DT-2606 Add endpoint to get recall NSI by nomsNumber

andymarke:
Url path indicates that the request is just for recalls, this is a departure from passing in the NSI codes required. This allows less knowledge of the internal Delius NSI codes


mikehalmamoj (approved):

lbennett-moj (commented):
LGTM

PR #477, opened 2021-06-10
DT-2606 Add attributes on Nsi when recall to indicate outcome and status

andymarke:
* Added isOutcomeRecall and isRecallRejectedOrWithdrawn on Nsi API 
* Only present when NSI is a recall

Added so that clients have a quick view of the recall without having to understand the meaning of Delius ref data codes


lbennett-moj (approved):
LGTM

PR #478, opened 2021-06-14
MAS-420 only attempt a tier writeback if offender is not soft deleted

markreesmoj:
And convert softDeleted on Offender to a boolean...


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Offender.java
```diff
@@ -101,7 +101,7 @@
     private String emailAddress;
 
     @Column(name = "SOFT_DELETED")
-    private Long softDeleted;
+    private boolean softDeleted;
```
andymarke, line 104: 👍🏿

andymarke (approved):
LGTM

andymarke (approved):
LGTM

PR #480, opened 2021-06-17
IC-1722 Always set RAR Activity Flag when contact is linked to a RAR Requirement

mahalrmoj:
## What does this pull request do?
Always set RAR Activity Flag when contact is linked to a RAR Requirement, even when it doesn't count towards RAR days (i.e. false), regardless of whether the contact has a type that is defined as "RAR Activity No".

## What is the intent behind these changes?
Correction to the logic otherwise Delius-Api rejects the contact.

## Breaking Changes?
None


axle-h (approved):

mxco86 (approved):
LGTM

PR #481, opened 2021-06-18
DT-2060 Recall Offender endpoint added

karenmillermoj:


src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
```diff
@@ -122,4 +123,16 @@ public Custody getCustodyByConvictionId(final @PathVariable String crn,
         return custodyService.getCustodyByConvictionId(crn, convictionId);
     }
 
+    @RequestMapping(value = "offenders/nomsNumber/{nomsNumber}/recall", method = RequestMethod.PUT, consumes = "application/json")
```
andymarke, line 126: very picky, but I think it should be `/recalled`   since essentially the recall has happened  rather than this is requesting a recall

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/CustodyResource.java
karenmillermoj: np - will update

PR #483, opened 2021-06-22
DT-2060 Store recalled date of offender

karenmillermoj:


src/main/java/uk/gov/justice/digital/delius/data/api/OffenderRecalledNotification.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.NoArgsConstructor;
+
+import java.time.LocalDate;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class OffenderRecalledNotification {
+    @ApiModelProperty(value = "The date the recall occurred", example = "2020-10-25")
```
andymarke, line 16: very minor but important, but I think this should be "The date the person was returned to custody"

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/OffenderRecalledNotification.java
karenmillermoj: Sure.  will update

PR #486, opened 2021-06-23
E2E-555: added court & responsible court to conviction dto

axle-h:
Adds new fields to the sentence dto.
* responsible court: mapped through from event.court (this is actually named responsibleCourt in delius)
* courtAppearance: latest sentence appearance or latest appearance otherwise. Logic taken from delius.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.delius.data.api;
 
+import io.swagger.annotations.ApiModelProperty;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
```diff
@@ -23,4 +23,10 @@
     private Sentence sentence;
     private KeyValue latestCourtAppearanceOutcome;
     private Custody custody;
+    private Court responsibleCourt;
```
andymarke, line 26: maybe take the opportunity to add a couple of Swagger API annotations on these two new fields and the latestCourtAppearanceOutcome field.

Sadly they have always been missing but given it isn't clear the difference it would be nice to add them.

andymarke (approved):
LGTM

andymarke (approved):
LGTM

PR #487, opened 2021-06-24
IC-1976 Added 'team email address' to /staff/username/{username} response and corrected exception type

mahalrmoj:
Refer and Monitor require a Team email address which has been added as part of this PR. 
Secondly, in production there are sentences with multiple RAR requirements which is not valid - 1% of sentences. Community-api now returns a http-400 in this case (rather than a 500).

Supply additional data to R&M and correct response codes. 

**Breaking Changes**
None. Additional attribute in Team.


axle-h (approved):

andymarke (approved):
LGTM

PR looks fine however out of 30,000 teams only 79 have email address so feels returning this doesn't add anything meaningful .  Is this for some future change is the client (interventions ??) hoping that it will start being used?

PR #489, opened 2021-06-25
DT-2239 Add nomsPrisonInstitutionCode to offender recalled endpoint

karenmillermoj:


src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -228,18 +228,22 @@ public Custody getCustodyByConvictionId(final String crn, final Long convictionI
...
+        log.info("Offender {} recalled from {} on {}", nomsNumber, recalledNotification.getNomsPrisonInstitutionCode(), recalledNotification.getRecallDate());
+        final var telemetryProperties = Map.of("offenderNo", nomsNumber,
+            "recallDate", recalledNotification.getRecallDate().format(DateTimeFormatter.ISO_DATE),
+            "institution", recalledNotification.getNomsPrisonInstitutionCode());
 
         try {
             Event event = convictionService.getActiveCustodialEvent(offender.getOffenderId());
-            telemetryClient.trackEvent( "P2POffenderRecalled");
+
+            telemetryClient.trackEvent("P2POffenderRecalled", telemetryProperties, null);
```
petergphillips, line 243: not sure if this telemetry looks correct - I thought prison to probation used `P2P` rather than community api?

petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
```diff
@@ -228,18 +228,22 @@ public Custody getCustodyByConvictionId(final String crn, final Long convictionI
...
             return ConvictionTransformer.custodyOf(event.getDisposal().getCustody());
```
petergphillips, line 244: new field only looks like it is used in logging?

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
karenmillermoj: I kinda followed the pattern that was already being used.  it uses the P2P prefix for those services currently called by P2P.

src/main/java/uk/gov/justice/digital/delius/service/CustodyService.java
karenmillermoj: Yes, it's a placeholder until the Delius team let us know what they want.

PR #493, opened 2021-06-28
✨ PIC-1417: Apply limited access markers to registration endpoint

jhevans:
This applies the same rules as are currently applied to the offender by crn endpoint to the registrations endpoint by crn.  

I had thought from previous checks in azure that court-case-service was the only consumer of this endpoint but having checked again for good measure it looks like this isn't the case so this will need some discussion with the maintainers of hmpps-tier and maybe others before merging.

As part of this I've also taken the liberty of removing the feature toggle on the UserAccessService as this has been turned on in live for a while now.


andymarke (approved):
LGTM

PR #498, opened 2021-07-01
IC-2016 Referral Send must exclude inactive RAR requirements

mahalrmoj:
Community Api should exclude inactive requirements when deriving the RAR requirement for a CRN sentence during the process of initiating a NSI in Delius. Also, any post request made for an NSI that has duplicates in Delius should result in a 400. The NSIs must be cleaned up in Delius before they can be acted upon.

Fix issues arising in production

**Breaking Changes**
None.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
@@ -116,7 +116,7 @@ public class RequirementService {
     // In the context of NSI's for interventions, the offender's sentence is the event against which
     // the NSI is created. This sentence is supplied by interventions as part of the "referral sent
     // request". This method takes the sentence and finds any associated requirements and returns one
-    // that has a main category of F - rehab activity requirement. It is invalid for multiple to exist.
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -119,10 +119,11 @@ public LicenceConditions getLicenceConditionsByConvictionId(String crn, Long con
     // that has a main category of F - rehab activity requirement. It is invalid for multiple to exist.
```
sldblog, line 119:
```suggestion
    // that has a main category of F - rehab activity requirement. It is invalid for multiple active ones to exist.
```

sldblog (commented):
LGTM, but will defer to code owners

src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
mahalrmoj: Done

electrified (dismissed):

electrified (approved):

PR #499, opened 2021-07-01
E2E-602: offender address type

axle-h:
New fields returned on offender address as in Delius:
* type: latest personal circumstances sub type
* type verified: latest personal circumstances evidenced flag
* latest assessment date: latest address assessment, assessment date
* created & updated datetimes


src/main/java/uk/gov/justice/digital/delius/data/api/Address.java
```diff
@@ -7,24 +7,64 @@
...
+    @ApiModelProperty(name = "Address type")
+    private KeyValue type;
+
+    @ApiModelProperty(name = "Address type is verified/evidenced", example = "true")
+    private Boolean typeVerified;
+
+    @ApiModelProperty(name = "If address previously assessed then the latest assessment date and time", example = "2021-06-11T12:00:00")
+    private LocalDateTime latestAssessmentDate;
+
+    @ApiModelProperty(name = "Date and time that this address was created", example = "2021-06-11T13:00:00")
```
andymarke, line 65: Curious, why do you need these?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/Address.java
axle-h: Not sure why the offender manager needs them and there are some technical issues with the placement on the mock-up here but they're part of the design. On second thoughts though the created time is not required currently - just thought them best to come as a pair?!
image

andymarke: cool, not an issue, just curious - rare that clients need these dates so nice to see use cases when they do

PR #500, opened 2021-07-02
E2E-602: expose new gender identity fields

axle-h:
New gender identity fields on offender now returned in offender detail response


andymarke (approved):
LGTM

PR #502, opened 2021-07-06
IC-2050 When creating an NSI and in the case of multiple RAR requirements, take the latest

mahalrmoj:
**What is included in this change?**
The creation of a NSI and the association of a referral was implemented on the basis that there is only ever one active requirement per service user sentence. This is not always the case. It has been agreed that should multiple exist, the one with the latest start date should be chosen and associated to the NSI. Where start date is the same, the later create date time is used.

This is to resolve a matter arising in production with approximately 1% of referrals raised by R&M.

**Are there any breaking changes?**
None, but introduced createDateTime in api class Requirement.


mxco86 (approved):
LGTM

andymarke (approved):
LGTM

PR #504, opened 2021-07-08
Pic 1452 secure offender court report

chris-faulkner-mojdt:
Takes the insecure endpoint at 
[(https://community-api-public.test.probation.service.justice.gov.uk/swagger-ui/index.html?urls.primaryName=NewTech%20Private%20APIs#/Offender%20Court%20Reports/getOffenderCourtReportByCrnAndCourtReportIdUsingGET)] and add it to a new secured `CourtReportResource`.

The endpoint will also apply the same user exclusion / restriction checks as found in `OffenderResource`.

This will be consumed initially by court-case-service and only a small subset of the fields found in the original controller will be required here so a smaller response objact is used.


src/main/java/uk/gov/justice/digital/delius/controller/secure/CourtReportResource.java
```diff
@@ -0,0 +1,65 @@
...
+        log.info("Call to getOffenderCourtReportByCrnAndCourtReportId");
+        return courtReportResponseEntityOf(offenderService.offenderIdOfCrn(crn), courtReportId);
+    }
+
+    private ResponseEntity<CourtReportMinimal> courtReportResponseEntityOf(Optional<Long> maybeOffenderId,  Long courtReportId) {
+        final var maybeCourtReport = maybeOffenderId.flatMap(offenderId -> courtReportService.courtReportMinimalFor(offenderId, courtReportId));
+        return maybeCourtReport.map(courtReport -> new ResponseEntity<>(courtReport, OK))
+                                .orElseGet(() -> new ResponseEntity<>(HttpStatus.NOT_FOUND));
+
+    }
```
andymarke, line 63: very minor, and just an observation, we tend to throw  a NotFoundException and not return  ResponseEntity, but to be honest not sure if I prefer the entity approach anyway, a bit more functional.

src/main/java/uk/gov/justice/digital/delius/service/CourtReportService.java
```diff
@@ -37,12 +38,21 @@ public CourtReportService(CourtReportRepository courtReportRepository) {
...
+    public Optional<CourtReportMinimal> courtReportMinimalFor(Long offenderId, Long courtReportId) {
+
+        return notDeletedCourtReportFor(offenderId, courtReportId)
+            .map(CourtReportTransformer::courtReportMinimalOf);
+    }
+
+    private Optional<uk.gov.justice.digital.delius.jpa.standard.entity.CourtReport> notDeletedCourtReportFor(Long offenderId, Long courtReportId) {
+        Optional<uk.gov.justice.digital.delius.jpa.standard.entity.CourtReport> maybeCourtReport = courtReportRepository.findByOffenderIdAndCourtReportId(offenderId, courtReportId);
+        return maybeCourtReport.filter(this::notDeleted);
+    }
```
andymarke, line 54: Observation, slightly nicer to move the deleted filter to repo, we have noticed a few occasions where bad/corrupt data in Delius is causing JPA errors, even for records we end up filtering, e.g where lazy loading isn't happening.

I think I have concluded that might be better to push filtering down to repo (define the soft_deleted as a boolean makes it even easier). I think there are some examples around

andymarke (dismissed):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/CourtReportResource.java
chris-faulkner-mojdt: I was really just copying the code that already existed here. Will apply as you suggest.

src/main/java/uk/gov/justice/digital/delius/service/CourtReportService.java
chris-faulkner-mojdt: OK - have seen     `Optional<Offender> findByCrnAndSoftDeletedFalse(String crn);` . Previously, I have also seen it done with a "Where" annotation on the entity itself. There's only one place something similar happens.

andymarke (approved):
LGTM

PR #505, opened 2021-07-09
IC-2067 Exclude soft deleted RARs during NSI Creation/Referral Sent process

mahalrmoj:
What is included in this change?
During the referral start/NSI creation request, a requirement is selected from the specified sentence. Soft deleted requirements must be excluded from the selection.

What is the intent behind these changes?
This is to resolve a matter arising in production by R&M.

Are there any breaking changes?
None, but introduced attribute softDeleted in api class Requirement.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
@@ -127,6 +127,7 @@ public class RequirementService {
         return getRequirementsByConvictionId(crn, eventId)
             .getRequirements().stream()
             .filter(Requirement::getActive)
+            .filter(requirement -> !requirement.getSoftDeleted())
             .filter(requirement ->
                 ofNullable(requirement.getRequirementTypeMainCategory())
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
```diff
@@ -127,6 +127,7 @@ public LicenceConditions getLicenceConditionsByConvictionId(String crn, Long con
         return getRequirementsByConvictionId(crn, eventId)
             .getRequirements().stream()
             .filter(Requirement::getActive)
+            .filter(requirement -> !requirement.getSoftDeleted())
```
axle-h, line 130: isnt this technically a breaking change?

axle-h (approved):

src/main/java/uk/gov/justice/digital/delius/service/RequirementService.java
mahalrmoj: This method is only used by our Referral Service that creates NSIs. I have done a find usages to confirm this.

PR #507, opened 2021-07-09
E2E-630: personal contact address

axle-h:
expose personal contact address

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/data/api/AddressSummary.java
+++ b/src/main/java/uk/gov/justice/digital/delius/data/api/AddressSummary.java
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.delius.data.api;
+
+import io.swagger.annotations.ApiModelProperty;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder(toBuilder = true)
+@AllArgsConstructor
+@NoArgsConstructor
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/AddressSummary.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder(toBuilder = true)
+@AllArgsConstructor
+@NoArgsConstructor
+public class AddressSummary {
```
andymarke, line 13: What is the rational for having this class alongside the existing API Address class?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/data/api/AddressSummary.java
axle-h: The existing address is an offender address, mapped from OFFENDER_ADDRESS & includes offender specific stuff like type, status, assessments (so should really have been named as such).

axle-h: I thought about renaming it but this would be a breaking change for any client using a generated client

andymarke: Perfect. Makes sense

PR #508, opened 2021-07-12
Probation-in-court pact starter

chris-faulkner-mojdt:
Adds PACT verification test which runs as part of validate-integration. Will fail if the API responses for various states do not match those expected by the consumer, court-case-service. In other projects we have put these into other workflows. Did not want to start adding extra "-P" parameters until we have decided how we might want to take this forward here. Also, based on this, we can publish the results of the verification to the PACT broker.


andymarke (approved):
LGTM

PR #512, opened 2021-07-14
✨ PIC-1527 add awaiting PSR to conviction

chris-faulkner-mojdt:
Adds a flag to the Conviction to indicate whether or not that conviction is awaiting a PSR.  It looks amongst the outcomes on the appearances, which are already fetched as part of another field.


andymarke (approved):
LGTM

PR #513, opened 2021-07-14
DT-2054 Exclude recall NSIs which is in PSS phase

andymarke:
Endpoint does not return recall NSIs after the licence period, however in live it was observed that some Events had no licence expiry date but had entered the PPS phase (so therefore had passed licence expiry) - e.g LED was just not present in Delius but obviously does exists in real world.

So if sentence has entered PSS no need to consult LED since by definition licence period has expired.


chris-faulkner-mojdt (approved):
LGTM

PR #514, opened 2021-07-15
E2E-634: fix offender summary not returning preferred name

axle-h:
Preferred name was added to the offender detail api but not the offender summary api


andymarke (approved):
LGTM

PR #517, opened 2021-07-20
E2E-654: added additional sentences to conviction apis

axle-h:
Additional sentences are associated to events (convictions) so exposed via root conviction object. ~~Also exposing conviction notes (these appear under sentence notes in Delius).~~


src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -96,9 +97,23 @@ public static Conviction convictionOf(Event event) {
                 .responsibleCourt(Optional.ofNullable(event.getCourt()).map(CourtTransformer::courtOf).orElse(null))
                 .courtAppearance(courtAppearances.map(CourtAppearanceBasicTransformer::latestOrSentencingCourtAppearanceOf).orElse(null))
                 .awaitingPsr(courtAppearances.map(CourtAppearanceBasicTransformer::awaitingPsrOf).orElse(false))
+                .notes(event.getNotes())
+                .additionalSentences(Optional.ofNullable(event.getAdditionalSentences())
+                    .map(x -> x.stream().map(ConvictionTransformer::additionalSentenceOf).collect(toList()))
```
andymarke, line 102: I wonder if this would be better placed inside `sentence`  ? Not sure interested in your thoughts

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
axle-h: I thought that too, but left it out as from the data is possible to not have a sentence (disposal) but still have additional sentences... Now I'm writing it it does sound a little silly. Shall I move it? Happy to keep it clean.

andymarke: Checked Delius, looks like UI (correctly) doesn't allow an additional sentence without a sentence/disposal even if database base model allows it. Maybe just keep it as it is. I'm 50/50

The community-api model isn't brilliant anyway so happy for you to move it or keep it, if you move it no doubt we will find that edge case where there is a missing DISPOSAL but an ADDITIONAL_SENTENCE!

axle-h: OK, I moved it into the sentence as it really doesn't make sense to have additional sentences without a sentence. I agree there may be some silly corner case but if there is then that data won't represent an "additional" sentence and is trivial to detect so there's an opportunity to map just that case into a more domain appropriate model.

andymarke (approved):
LGTM

PR #521, opened 2021-07-29
Added contactDate on api contact

dnyanesh-sonwane:


electrified:
The date is included in the start and end fields. We've intentionally not exposed the date separately as it's a bit strange from an API design point of view, and is more an artifact of the way the database schema was designed

mikehalmamoj (approved):

dnyanesh-sonwane:
Will create a new one

PR #522, opened 2021-07-29
changed the endpoint to return contactsummary

dnyanesh-sonwane:


dnyanesh-sonwane:
This endpoint was only created yesterday evening, so no risk of breaking any existing  consumers.

mikehalmamoj (approved):

PR #524, opened 2021-08-02
IC-2141 Exclude "soft-deleted" nsi's

mahalrmoj:
As part of the integration with R&M, when finding a matching NSI Community-api must exclude "soft-deleted" NSI's.

This is to fix a rare situation after which duplicate NSIs are present in nDelius, which then need to be manually deleted from nDelius.

**Any breaking changes?**
None

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
+++ b/src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
@@ -115,6 +115,7 @@ public class ReferralService {
         var existingNsis = nsiService.getNsiByCodes(offenderId, sentenceId, Collections.singletonList(getNsiType(nsiMapping, contractType)))
             .map(wrapper -> wrapper.getNsis().stream()
                 // eventID, offenderID, nsi type are handled in the NSI service
+                .filter(nsi -> !nsi.getSoftDeleted())
                 .filter(nsi -> ofNullable(nsi.getReferralDate()).map(n -> n.equals(toLondonLocalDate(startedAt))).orElse(false))
                 .filter(nsi -> ofNullable(nsi.getNsiStatus()).map(n -> n.getCode().equals(nsiMapping.getNsiStatus())).orElse(false))
                 .filter(nsi -> Objects.isNull(nsi.getNsiOutcome()))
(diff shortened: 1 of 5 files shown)


electrified (approved):

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
```diff
@@ -38,6 +38,7 @@
                 .notes(n.getNotes())
                 .intendedProvider(ProbationAreaTransformer.probationAreaOf(n.getIntendedProvider(), INCLUDE_INTENDED_PROVIDER_TEAMS))
                 .active(zeroOneToBoolean(nsi.getActiveFlag()))
+                .softDeleted(zeroOneToBoolean(nsi.getSoftDeleted()))
```
andymarke, line 41: won't this always be false due to the above change in the filter?

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
```diff
@@ -115,6 +115,7 @@ public ReferralEndResponse endNsiReferral(final String crn, final String context
         var existingNsis = nsiService.getNsiByCodes(offenderId, sentenceId, Collections.singletonList(getNsiType(nsiMapping, contractType)))
             .map(wrapper -> wrapper.getNsis().stream()
                 // eventID, offenderID, nsi type are handled in the NSI service
+                .filter(nsi -> !nsi.getSoftDeleted())
```
andymarke, line 118: I like to suggest a breaking change; move the filter into the service call. Check with @chris-faulkner-mojdt that the endpoint added last year doesn't need deleted NSI. I would be 99% certain it does not and is an existing bug in that service.

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/transformers/NsiTransformer.java
mahalrmoj: Filtering on nsi's happen after transformation to api.Nsi

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
mahalrmoj: Will do

mahalrmoj:
Demo'd to R&M team and Matt

src/main/java/uk/gov/justice/digital/delius/service/ReferralService.java
mahalrmoj: Decided to proceed and will do this as a secondary change on Chris' approval

PR #527, opened 2021-08-03
E2E-635: add FTC count & breach end to conviction APIs

axle-h:


src/main/java/uk/gov/justice/digital/delius/data/api/Conviction.java
```diff
@@ -26,6 +26,12 @@
     @ApiModelProperty(name = "Conviction in breach flag", example = "true")
     private Boolean inBreach;
 
+    @ApiModelProperty(name = "Total number of failure to comply outcomes since the last breach end", example = "3")
+    private Long ftcCount;
```
andymarke, line 30: should this be `failureToComplyCount` rather than the initials? Not sure if "ftcCount" is a common enough acronym?

I'll leave this up to you to decide given I don't really know this part of the domain.

(BTW, internally it is fine, just the public API might need thinking about)

andymarke (approved):
LGTM

andymarke (approved):
LGTM

PR #531, opened 2021-08-06
IC-2165 New endpoint for only changing the location of an appointment

mahalrmoj:
Added a new community-api endpoint for relocation of an existing appointment where location is the only thing that has changed.

To avoid using the reschedule endpoint. The endpoint makes the original rescheduled time slot unavailable for any new appointments which would means a second change of location not possible even when the times remain the same.

**Any breaking changes?**
None

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
@@ -18,10 +18,12 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.justice.digital.delius.controller.advice.ErrorResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.AppointmentCreateResponse;
+import uk.gov.justice.digital.delius.data.api.AppointmentRelocateResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentRescheduleResponse;
 import uk.gov.justice.digital.delius.data.api.AppointmentUpdateResponse;
 import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentCreateRequest;
 import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentOutcomeRequest;
+import uk.gov.justice.digital.delius.data.api.AppointmentRelocateRequest;
 import uk.gov.justice.digital.delius.data.api.ContextlessAppointmentRescheduleRequest;
 import uk.gov.justice.digital.delius.service.AppointmentService;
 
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentRelocateResponse.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.validation.constraints.NotNull;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class AppointmentRelocateResponse {
+
+    @NotNull
+    private Long appointmentId;
```
electrified, line 17: The response is a bit redundant as you get back the appointment ID originally passed in. Is there anything useful we could return?

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
```diff
@@ -99,6 +101,25 @@ public AppointmentRescheduleResponse rescheduleAppointmentWithContextName(final
...
     }
 
+    @RequestMapping(value = "/offenders/crn/{crn}/appointments/{appointmentId}/relocate",
+        method = RequestMethod.POST,
+        consumes = "application/json")
+    @ApiResponses(
+        value = {
+            @ApiResponse(code = 200, message = "Updated", response = String.class),
+            @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+            @ApiResponse(code = 403, message = "Requires role ROLE_COMMUNITY_INTERVENTIONS_UPDATE"),
```
electrified, line 111: Perhaps this could have a less specific role in the future as it's not particularly interventions bespoke, and could maybe more widely used, unlike some of the other endpoints in this controller.

electrified (approved):

src/main/java/uk/gov/justice/digital/delius/data/api/AppointmentRelocateResponse.java
mahalrmoj: I see it as a placeholder for attributes to be added as and when required

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
mahalrmoj: @mxco86 do you have a feel for what other roles could be added to make this less specific to interventions?

mxco86: If and endpoint is more general it usually requires 'ROLE_COMMUNITY'

mxco86 (dismissed):

src/main/java/uk/gov/justice/digital/delius/controller/secure/AppointmentBookingController.java
andymarke: I suggest leaving ROLE_COMMUNITY for read-only endpoints. Rightly or wrongly (wrongly IMHO) we are not using scope - e.g read,write - to control access to our endpoints, just role.

Given that maybe something like `ROLE_COMMUITY_APPOINTMENTS_UPDATE` so the update endpoints are at least locked to certain clients. For now, give everything else is `ROLE_COMMUNITY_INTERVENTIONS_UPDATE` maybe it doesn't matter until smoother client needs to use this endpoint.

src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
```diff
@@ -34,6 +35,15 @@ public static JsonPatch mapAttendanceFieldsToOutcomeOf(final ContextlessAppointm
...
     }
 
+    public static JsonPatch mapOfficeLocation(final String officeLocation) {
+
+        final var patchOperations = new ArrayList<JsonPatchOperation>();
+
+        patchOperations.add(new ReplaceOperation(of(TARGET_OFFICE_LOCATION_FIELD_NAME), valueOf(officeLocation)));
+
+        return new JsonPatch(patchOperations);
+    }
```
andymarke, line 45: I'm guessing you could just write this as `return new JsonPatch(List.of(new ReplaceOperation(of(TARGET_OFFICE_LOCATION_FIELD_NAME), valueOf(officeLocation)))))`  ?

andymarke (approved):
LGTM

mxco86 (approved):
LGTM

PR #541, opened 2021-08-31
E2E-730 - Add include filter to contacts endpoint. Previously it was …

electrified:
…not possible to return contacts of a particular type AND appointments. The new include qurey parameter allows multiple filters to be added together. In the future this could be potentially expanded to include a category filter

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
@@ -266,7 +266,8 @@ public class OffendersResource {
         final @RequestParam(value = "attended", required = false) Optional<Boolean> attended,
         final @RequestParam(value = "complied", required = false) Optional<Boolean> complied,
         final @RequestParam(value = "nationalStandard", required = false) Optional<Boolean> nationalStandard,
-        final @RequestParam(value = "outcome", required = false) Optional<Boolean> outcome) {
+        final @RequestParam(value = "outcome", required = false) Optional<Boolean> outcome,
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
```diff
@@ -266,7 +266,8 @@ public OffenderDetailSummary getOffenderDetails(
         final @RequestParam(value = "attended", required = false) Optional<Boolean> attended,
         final @RequestParam(value = "complied", required = false) Optional<Boolean> complied,
         final @RequestParam(value = "nationalStandard", required = false) Optional<Boolean> nationalStandard,
-        final @RequestParam(value = "outcome", required = false) Optional<Boolean> outcome) {
+        final @RequestParam(value = "outcome", required = false) Optional<Boolean> outcome,
+        final @ApiParam(name = "include", value = "Contacts to include. Can be a contact type code, prefixed with 'type_' or appointments", example = "type_ccmp", allowMultiple = true) @RequestParam(value = "include", required = false) Optional<List<String>> include) {
```
andymarke, line 270: Could this just be a List and treat an empty list as not present ?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java
axle-h: @andymarke looks like the existing contactTypes was used as a template for this new param so although an empty list would work fine, it would be inconsistent.

PR #544, opened 2021-09-06
IC-2207 - For all unattended appointments a Refer to Office Manager alert is now requested

mahalrmoj:
Prior to this change a "refer to office manager" alert was only raised when the offender's behaviour was bad. After this change, the alert is also raised for all non attendance, both supplier assessments and service delivery.

In using R&M, Probation Practitioners were also expecting a ROM alert for non-attendance which was highlighted during feedback sessions.

**Are there any breaking changes?**
None

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
+++ b/src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
@@ -18,6 +18,7 @@ import static java.util.Optional.ofNullable;
 @AllArgsConstructor
 public class AppointmentPatchRequestTransformer {
 
+    private static final String NON_ATTENDANCE_VALUE = "no";
     private static final String TARGET_NOTES_FIELD_NAME = "notes";
     private static final String TARGET_OUTCOME_FIELD_NAME = "outcome";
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/digital/delius/transformers/AppointmentPatchRequestTransformer.java
```diff
@@ -55,7 +56,7 @@ private static void addReplaceOperationForOutcomeIfAttended(final IntegrationCon
 
         patchOperations.add(new ReplaceOperation(of(TARGET_OUTCOME_FIELD_NAME), valueOf(outcomeType)));
 
-        if ( notifyBehaviour ) {
+        if ( notifyBehaviour || NON_ATTENDANCE_VALUE.equalsIgnoreCase(attended)) {
```
mxco86, line 59: If this logic to notify on non-attended appointments applies to all appointments should the logic go into the Delius API?

mahalrmoj: Does the "Refer to Office Manager" alert apply to all clients where an appointment id non-attended? I don't know how to answer this question. I will put this question to Andrew

PR #545, opened 2021-09-06
E2E-735 Add cases endpoint

electrified:
Adds a new "Cases" endpoint initially for the use of the Manage a Supervision pilot service.


src/main/java/uk/gov/justice/digital/delius/controller/secure/StaffResource.java
```diff
@@ -79,4 +84,18 @@ public StaffDetails getStaffDetailsForUsername(
...
+        @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+        @ApiResponse(code = 404, message = "Not found", response = ErrorResponse.class),
+        @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)})
+    @GetMapping(path = "/staff/username/{username}/cases")
+    public List<Case> getCases(@ApiParam(name = "username", value = "Delius username", example = "SheliaHancockNPS", required = true)
+                                                           @NotNull
+                                                           @PathVariable(value = "username") final String username,
+                                                 @PageableDefault(sort = {"secondName", "firstName"}, direction = Direction.ASC, size = Integer.MAX_VALUE) final Pageable pageable) {
+        return staffService.getOffenderCasesForUser(username, pageable);
+    }
```
axle-h, line 100: shouldnt this return a Page<Case> not a List<Case> ?

axle-h:
Given that this is currently our exclusive domain, the naming of `Case` does seem a bit pre-emptive. Maybe something that ties it to the API like `StaffCaseloadEntry`.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepository.java
```diff
@@ -42,6 +43,21 @@ public DuplicateOffenderException(String message) {
...
+        "where o.softDeleted != 1 " +
+        "and o.offenderId in (" +
+            "select e.offenderId from Event e " +
+            "join e.disposal d join d.requirements r " +
+            "join r.requirementTypeMainCategory rtmc " +
+            "where e.softDeleted != 1 and e.activeFlag = 1 and d.softDeleted != 1 and r.startDate < CURRENT_TIMESTAMP AND (r.terminationDate IS NULL OR r.terminationDate > CURRENT_TIMESTAMP) and rtmc.code = 'F' " +
+            "group by e.offenderId " +
+            "having count(e.eventId) = 1) " +
+        "and u.distinguishedName = :username")
+    List<Offender> getOffendersWithOneActiveEventAndRarRequirementForStaff(@Param("username") String username, Pageable pageable);
```
axle-h, line 59: this isnt filtering to community orders only?

src/main/java/uk/gov/justice/digital/delius/controller/secure/StaffResource.java
```diff
@@ -79,4 +84,18 @@ public StaffDetails getStaffDetailsForUsername(
...
+        @ApiResponse(code = 400, message = "Invalid request", response = ErrorResponse.class),
+        @ApiResponse(code = 404, message = "Not found", response = ErrorResponse.class),
+        @ApiResponse(code = 500, message = "Unrecoverable error whilst processing request.", response = ErrorResponse.class)})
+    @GetMapping(path = "/staff/username/{username}/cases")
+    public Page<StaffCaseloadEntry> getCases(@ApiParam(name = "username", value = "Delius username", example = "SheliaHancockNPS", required = true)
+                                                           @NotNull
+                                                           @PathVariable(value = "username") final String username,
+                                             @PageableDefault(sort = {"secondName", "firstName"}, direction = Direction.ASC, size = Integer.MAX_VALUE) final Pageable pageable) {
+        return staffService.getOffenderCasesForUser(username, pageable);
+    }
```
andymarke, line 100: Small bit of caution: Delius *optionally* maps a Delius user and staff account. But that mapping is not always present and we have found a significant number where that linking hasn't happened.
The HDC licences project has been busy fixing this as it roles out region by region, and each region it roles out to we might find 20 or 30 staff accounts not linked.

Therefore this won't always return results as expected

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepository.java
```diff
@@ -42,6 +44,25 @@ public DuplicateOffenderException(String message) {
...
+            "join d.requirements r  " +
+            "join r.requirementTypeMainCategory rtmc " +
+            "where e.softDeleted != 1 and e.activeFlag = 1 and d.softDeleted != 1 " +
+            "and r.startDate < CURRENT_TIMESTAMP AND (r.terminationDate IS NULL OR r.terminationDate > CURRENT_TIMESTAMP) " +
+            "and dt.sentenceType = 'SP' and rtmc.code = 'F' " +
+            "group by e.offenderId " +
+            "having count(e.eventId) = 1) " +
+        "and u.distinguishedName = :username")
+    Page<Offender> getOffendersWithOneActiveEventCommunitySentenceAndRarRequirementForStaff(@Param("username") String username, Pageable pageable);
+
```
andymarke, line 65: a future change might be to make the endpoint generic and pass in a filter, so by default it would return the entire case load but you could set a filter with 
singleConviction = true
rarRequirement = true

etc etc - and then use a JPA predicate to build up the query.

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/controller/secure/StaffResource.java
electrified: Thanks @andymarke. In these cases how is the staff id -> user mapping resolved. Is it a manual process or is there an alternative way to join tables to determine the staff_id from the username?

PR #548, opened 2021-09-10
E2E-735 Fix username case sensitivity on cases endpoint

electrified:


petergphillips (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepository.java
```diff
@@ -60,7 +60,7 @@ public DuplicateOffenderException(String message) {
             "and dt.sentenceType = 'SP' and rtmc.code = 'F' " +
             "group by e.offenderId " +
             "having count(e.eventId) = 1) " +
-        "and u.distinguishedName = :username")
+        "and upper(u.distinguishedName) = upper(:username)")
```
petergphillips, line 63: watch out for performance - using an `upper` I think will mean that the query won't use the `distinguishedName` index so could be slow...

andymarke: @electrified where does the username come from to be supplied to endpoint? Does just come from the database anyway? Or does someone type it in?

andymarke (approved):
LGTM

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/OffenderRepository.java
electrified: @andymarke It comes from the HMPPS auth issued JWT. However in the case of at least one user, what we get back from the JWT is all caps, and the Delius DB username is mixed case, so needs to be case insensitive.

petergphillips: Interesting.  The NOMIS username is all caps so it is probably a hangup from that time in that we deliberately uppercase usernames.  Perhaps should look at not changing the case of the username when putting into the jwt...

PR #554, opened 2021-09-24
E2E-794: experimental activity log API

axle-h:
Experimental activity log API. Some differences from the existing contact summary API that has driven this approach:
* Pagination by contact date
* Ability to count RAR days from source contacts - cached values on NSI & Requirement are known to have issues & Delius actually uses a stored procedure to render RAR count on a RAR requirement page... this API is our alternative, with which we have a way to paginate RAR activity and an efficient method for counting it
* Offender level contacts by conviction dates - we're looking for all 'activity' across the lifetime of a conviction (conviction date to sentence termination date) and it is known that some relevant contacts are not associated to the event directly.
* More RAR details required including joining on the NSI & requirement

image


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactDateRepositoryImpl.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.persistence.PersistenceContext;
+import java.time.LocalDate;
+
+import static org.springframework.data.jpa.repository.query.QueryUtils.toOrders;
+
+/**
+ * Repository of distinct contact dates.
+ */
+@Repository
+public class ContactDateRepositoryImpl implements ContactDateRepository {
```
andymarke, line 22: I guess the below couldn't be written as a JPA query since you need the specification has a starting point.

I was wondering if PagingAndSortingRepository would help here at all, but can't think how

andymarke (approved):
LGTM

I tend to pass nulls in to functions and convert to Optional when required, but that a personal style thing rather than "this is how it would be done"

Shame that can't think of a better solution for the distinct data stuff, but looks like you have the best solution.

But all looks good to me

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactDateRepositoryImpl.java
axle-h: Yeah so I did a fair bit of investigating on this as I also wasn't happy with going bare JPA. As you say the problem is that there is no way to build JPQL and then apply a Specification and there's no overload of `findAll` that takes a Specification whilst providing some way of modifying the select query. Key for this repository is `distinct CONTACT_DATE`. It is I admit an unusual case and the resulting raw JPA is pretty compact so in the end I was happy to leave it in.

andymarke (approved):
LGTM

PR #558, opened 2021-10-04
Trello 796 surface enforcement action on activity log

dnyanesh-sonwane:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
```diff
@@ -187,6 +193,10 @@
     @Column(name = "RAR_ACTIVITY")
     private String rarActivity;
 
+    @OneToMany(cascade = CascadeType.ALL, mappedBy = "contact", orphanRemoval = true)
+    @Default
+    private List<Enforcement> enforcements = new ArrayList<>();
```
axle-h, line 198: I presume the default here is a copy-paste from delius api? I dont think we need it here as that was about nullability in kotlin & defaults when creating new contacts

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
```diff
@@ -0,0 +1,87 @@
...
+    @LastModifiedDate
+    private LocalDateTime lastupdatedDateTime;
+
+    @Column(name = "CREATED_BY_USER_ID", nullable = false)
+    @Default
+    private Long createdByUserId = 0L;
+
+    @Column(name = "LAST_UPDATED_USER_ID", nullable = false)
+    @Default
+    private Long lastUpdatedUserId = 0L;
```
axle-h, line 81: are all these defaults needed?

src/main/java/uk/gov/justice/digital/delius/data/api/Enforcement.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Enforcement {
+    @NotNull
+    @ApiModelProperty(name = "EnforcementAction", required = true)
+    private KeyValue enforcementAction;
+}
```
axle-h, line 19: is this really all we need here? none of the dates?

src/main/java/uk/gov/justice/digital/delius/data/api/Enforcement.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotNull;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Enforcement {
+    @NotNull
```
andymarke, line 16: maybe "required" should be force, it has no real impact but people might think it means there is always an enforcement

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
```diff
@@ -211,12 +221,20 @@ public boolean isRarActivity() {
 
         // the NSI is the primary component of RAR, so we check that first.
         // we also don't have to be careful with defaults here as the requirement and nsi foreign keys are mutually exclusive.
-        return Optional.ofNullable(getNsi())
+        return ofNullable(getNsi())
```
andymarke, line 224: Very minor, personally I prefer not to statically import Optional, a bit like if you have `List.of` it would be weird to just have `of("banana")`

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
```diff
@@ -211,12 +221,20 @@ public boolean isRarActivity() {
...
             .map(Either::<Nsi, Requirement>left)
-            .or(() -> Optional.ofNullable(getRequirement())
+            .or(() -> ofNullable(getRequirement())
                 .filter(r -> !r.getSoftDeleted() && r.isRarRequirement())
                 .map(Either::<Nsi, Requirement>right))
             .orElse(null);
     }
+
+    public Optional<Enforcement> getEnforcement() {
+        if (enforcements.size() > 1) {
```
andymarke, line 234: the data model allows multiple, are you 100% sure in live we never have more than 1. If there is won't this mean the API will fail?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
```diff
@@ -0,0 +1,87 @@
...
+import static javax.persistence.GenerationType.AUTO;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder(toBuilder = true)
+@EntityListeners(AuditingEntityListener.class)
+@Entity
+@Table(name = "ENFORCEMENT")
+@Where(clause = "SOFT_DELETED = 0")
```
andymarke, line 37: 👍🏿

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
```diff
@@ -0,0 +1,87 @@
...
+    @Column(name = "SOFT_DELETED", nullable = false)
+    @Default
+    private Boolean softDeleted = false;
+
+    @Column(name = "ROW_VERSION", nullable = false)
+    @Version
+    @Default
+    private Long rowVersion = 1L;
+
+    @Column(name = "LAST_UPDATED_DATETIME")
```
andymarke, line 71: You could inherit from AuditableEntity and drop this attributes

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
andymarke: feels like an unnecessary  Kotlin cut & paste  😄

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/EnforcementAction.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.persistence.Table;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder(toBuilder = true)
+@EntityListeners(AuditingEntityListener.class)
+@Entity
+@Table(name = "R_ENFORCEMENT_ACTION")
+@Where(clause = "SELECTABLE = 'Y'")
```
andymarke, line 26: Hmm, not sure about this. I think SELECTABLE means "is this enforcement available in the UI" which is different to what you need. If a contact is related to a enforcement action that has been deprecated then it should still be returned right?

andymarke (approved):
LGTM

But I would look at whether Contact to Enforcement is One to One or One to Many in real data. Check preprod database

Also I think the filter on Enforcement action should be removed

src/main/java/uk/gov/justice/digital/delius/data/api/Enforcement.java
dnyanesh-sonwane: We are only showing enforcement action description, so don't need anything else.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
dnyanesh-sonwane: Fixed

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
dnyanesh-sonwane: Although data model allows this, in reality we would never have more than one enforcements for a contact, hence this validation. Confirmed by @axle-h

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
dnyanesh-sonwane: Done

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/EnforcementAction.java
dnyanesh-sonwane: Yes, thats correct. Fixed

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
dnyanesh-sonwane: Fixed

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
dnyanesh-sonwane: Fixed

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
andymarke: OK, given that then just map it as a one-to-one

`
    @OneToOne(mappedBy = "contact")
     private Enforcement enforcement;
`

src/main/java/uk/gov/justice/digital/delius/data/api/Enforcement.java
dnyanesh-sonwane: EnforcementAction is required if there is an Enforcement, hence required true

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Contact.java
dnyanesh-sonwane: Done

PR #562, opened 2021-10-12
E2E-885 - Enhance appointment types metadata endpoint

electrified:
Add additional metadata to the response, so it can be determined what the appointment can be booked against (the offender, events, requirement types)

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactTypeRepository.java
+++ b/src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactTypeRepository.java
@@ -10,13 +10,11 @@ import java.util.Optional;
 public interface ContactTypeRepository extends JpaRepository<ContactType, Long> {
     Optional<ContactType> findByCode(String code);
 
-    /*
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactTypeRepository.java
```diff
@@ -10,13 +10,11 @@
...
     Optional<ContactType> findByCode(String code);
 
-    /*
-        Category code "AL" refers to the "All" category.
-    */
-    @Query("SELECT DISTINCT type FROM ContactType type " +
-        "INNER JOIN FETCH type.contactCategories category " +
-        "WHERE category.codeValue = 'AL' AND type.selectable = true AND type.scheduleFutureAppointments = 'Y' " +
-        "AND type.attendanceContact = true")
+    @Query("""
```
electrified, line 13: Filtering to the "All" category and those with scheduleFutureAppointments was unnecessarily restrictive, as those types are still available in Delius.

PR #565, opened 2021-10-18
E2E-839 - Add provider (probation area) to staff response

electrified:


src/main/java/uk/gov/justice/digital/delius/data/api/StaffDetails.java
```diff
@@ -28,4 +28,6 @@
     private Human staff;
     @ApiModelProperty(value = "all teams related to this staff member")
     private List<Team> teams;
+    @ApiModelProperty(value = "provider this staff member is associated with")
+    private ProbationArea probationArea;
```
electrified, line 32: Our current need is just for the code and description, so contemplated adding as a KeyValue to make the change to the response as minimal as possible.

andymarke: yeah, and even though I can see you don't fill in the entire ProbationArea object it is a beast

PR #572, opened 2021-11-03
Trello-166 R&M CRSNOTE notification requests are idempotent

mahalrmoj:
https://trello.com/c/dbIMCNNo/166-make-notification-requests-idempotent

The notification service creates contacts in nDelius for the event specified in the request. This change makes that request idempotent, using the following attributes to determine if the contact already exists for commission-rehabilitation-services:

- Offender Id
- NSI Id
- Contact Type (In this case CRSNOTE)
- Contact Date
- Contact Time

This key may not be applicable to all client contexts. This key is defined in the code to apply to CRS Contacts.

The intent is to make the link between R&M and NDelius more robust and remove the opportunity for duplicates to arise.

**Any breaking changes?**
There are no breaking API changes as such. However the call is now idempotent.


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ContactRepository.java
```diff
@@ -48,4 +48,17 @@
     Optional<Contact> findByContactIdAndOffenderIdAndContactTypeAttendanceContactIsTrueAndSoftDeletedIsFalse(Long contactId, Long offenderId);
 
     Optional<Contact> findByContactIdAndOffenderIdAndSoftDeletedIsFalse(Long contactId, Long offenderId);
+
+    @Query("SELECT contact FROM Contact contact "
```
electrified, line 52: This multi line string could be a Text Block

src/main/java/uk/gov/justice/digital/delius/service/NotificationService.java
```diff
@@ -33,10 +42,16 @@ public NotificationResponse notifyContact(final String crn, final Long sentenceI
 
         assertAppointmentType(request.getContactType());
 
-        final var newContact = makeNewContact(crn, sentenceId, request);
-        final var contactDto = deliusApiClient.createNewContact(newContact);
-
-        return NotificationResponse.builder().contactId(contactDto.getId()).build();
+        return matchExistingContact(crn, request)
+            .map(contactId -> {
+                log.debug(format("Notification already exists for CRN: %s, Sentence: %d, ContactType: %s and ContactDateTime: %s", crn, sentenceId, request.getContactType(), request.getContactDateTime()));
```
electrified, line 47: This log message doesn't include everything used by the DB search. e.g. should it log NSI ID too?

src/main/java/uk/gov/justice/digital/delius/service/NotificationService.java
```diff
@@ -59,6 +74,22 @@ private void assertAppointmentType(String contactTypeCode) {
         }
     }
 
+    private Optional<Long> matchExistingContact(String crn, NotificationCreateRequest request) {
+        Offender offender = offenderRepository.findByCrn(crn).orElseThrow(() -> new BadRequestException(format("Cannot find offender with CRN: %s", crn)));
+        var contactLocalDate = toLondonLocalDate(request.getContactDateTime());
+        var contactStartLocalTime = toLondonLocalTime(request.getContactDateTime()).truncatedTo(ChronoUnit.SECONDS);
+
+        return contactRepository.findByOffenderAndNsiIdAndContactTypeAndContactDateAndSoftDeletedIsFalse(
```
electrified, line 82: Should this match on the requirement ID as that is a parameter too?

What about the note text/ team / staff member?

src/main/java/uk/gov/justice/digital/delius/service/NotificationService.java
mahalrmoj: Yeah will log it and perhaps as a WARN to so we can see how often it arises

src/main/java/uk/gov/justice/digital/delius/service/NotificationService.java
mahalrmoj: Will discuss with @mxco86 but the basic reasoning was that within an NSI there should never be a duplicate R&M event for the offender at any point in time.

mahalrmoj: Conclusion is that we are going to make the method CRS specific and in that context the current equality check is adequate. The naming will make it clear.

electrified (approved):

PR #579, opened 2021-12-01
WFP-755 added grade to offender manager

carlov20:


src/main/java/uk/gov/justice/digital/delius/transformers/OffenderManagerTransformer.java
```diff
@@ -111,4 +113,13 @@ private static boolean isUnallocated(final PrisonOffenderManager offenderManager
                 .orElse(false);
     }
 
+    private static OffenderManagerGrade offenderManagerGradeOf(final OffenderManager offenderManager) {
+        return Optional.ofNullable(offenderManager.getOfficer().getGrade())
+            .map(grade -> OffenderManagerGrade.builder()
+                .code(grade.getCodeValue())
+                .description(grade.getCodeDescription())
+                .build())
+            .orElse(null);
```
trozzomoj, line 122: I do not really understand the use of orEsle(null) , that I can see is used everywhere , personally I believe is a bad practice and should be replaced with Optional.orElse()

trozzomoj (dismissed):

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Officer.java
```diff
@@ -36,4 +39,8 @@
 
     @Column(name = "FORENAME2")
     private String forename2;
+
+    @ManyToOne
+    @JoinColumn(name = "STAFF_GRADE_ID")
```
markreesmoj, line 44: I am very scared of this DB
(fetch = FetchType.LAZY)
might be an idea here?
But ask someone who knows what they are talking about

src/main/java/uk/gov/justice/digital/delius/data/api/OffenderManagerGrade.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class OffenderManagerGrade {
```
andymarke, line 13: This works fine but there is a generic KeyValue class if you prefer to use that

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_GetAllOffenderManagersAPITest.java
```diff
@@ -132,6 +135,9 @@ void canGetAllOffenderManagersByCrn() {
             assertThat(communityOffenderManager.getTeam()).isNotNull();
             assertThat(communityOffenderManager.getStaffCode()).isEqualTo("SH0007");
             assertThat(communityOffenderManager.getFromDate()).isEqualTo(LocalDate.of(2018, Month.MAY, 4));
+            assertThat(communityOffenderManager.getGrade()).isNotNull();
+            assertThat(communityOffenderManager.getGrade().getDescription()).isEqualTo("Accepted");
```
andymarke, line 139: Hmm, this looks like an odd grade, what has happened here?

andymarke (dismissed):
LGTM

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_GetAllOffenderManagersAPITest.java
carlov20: Its how the officer was created in the db https://github.com/ministryofjustice/community-api/blob/main/src/main/resources/db/data/V1_4__offenders_licences_data.sql#L250 where their STAFF_GRADE_ID was set to the reference number 1, happy to add in a script which sets the officer to an actual grade and assert on that?

andymarke (approved):
LGTM

PR #586, opened 2021-12-22
Add NSI mappings for Greater Manchester services

sldblog:
As defined in https://github.com/ministryofjustice/hmpps-interventions-service/pull/774


mxco86 (approved):
LGTM

PR #587, opened 2022-01-10
PI-5: Update H2 CVE-2021-42392 & CVE-2021-23463

axle-h:
H2 upgrade breaking changes dealt with:

* `ARRAY` columns must be typed
* `VALUE` is now a reserved word
* `CHAR()` fields are right-padded with spaces
* Oracle dialect is now more strict e.g. the `LIMIT` clause is not allowed


src/main/java/uk/gov/justice/digital/delius/config/NationalDatasourceConfig.java
```diff
@@ -37,8 +37,9 @@ public LocalContainerEntityManagerFactoryBean nationalEntityManager(@Qualifier("
         HashMap<String, Object> properties = new HashMap<>();
         properties.put("hibernate.hbm2ddl.auto",
                 env.getProperty("hibernate.hbm2ddl.auto"));
+        final var dialect = env.getProperty("hibernate.dialect");
```
marcus-bcl, line 40: Note `getProperty` can take a default value, to save doing the null-check yourself e.g.
```suggestion
        final var dialect = env.getProperty("hibernate.dialect", "org.hibernate.dialect.Oracle12cDialect");
```

marcus-bcl (approved):

axle-h:
@marcus-bcl fixed the default value

marcus-bcl (approved):

PR #590, opened 2022-01-25
CVSL-334: Initial suggestions for team view of managed/unallocated cases.

timharrison-moj:
Implements controller & part of the service layer where it currently returns a stubbed response.

The presentation of results is very "flat" at present and I'm not sure this fits with the rest of the service? Quite happy to alter this so we get more structure in responses, for example, attributes for probationArea (code, desc), PDU (code, desc), LDU (code, desc), Team (id, code, desc), Staff (details), Offender (details) details separated out within the list of response objects?

I don't really see a need for the `current` parameter unless you have other needs where this might be useful? For CVL we're only interested in the current view - as it stands now, and which offenders are allocated or unallocated within the team.

The next challenge is to understand where this data will come from in the repository layer, which might then inform its structure a bit more. I know the caseload table was mentioned? 

I've left the tests until the intended structure of data is better understood, so it's just a starter for ten.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/delius/controller/secure/TeamResource.java
+++ b/src/main/java/uk/gov/justice/digital/delius/controller/secure/TeamResource.java
@@ -80,7 +80,7 @@ public class TeamResource {
     @PreAuthorize("hasRole('ROLE_COMMUNITY')")
     @ApiResponses(
         value = {
-            @ApiResponse(code = 200, message = "All active office locations for the specified team", response = OfficeLocation.class, responseContainer = "List"),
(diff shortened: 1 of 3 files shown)


mxco86 (commented):
I think a structured response would work, as long as it doesn't bloat things too much. It looks like you could make use of some existing data objects and that may cover everything?
```
ProbationArea
LDU
Team
OffenderDetailSummary
Staff
```

mxco86 (commented):
No need for a `current` parameter if we're not going to use it. I guess we should make it clear that the endpoint only returns currently managed people, either in the path or the documentation

marcus-bcl:
Superseded by https://github.com/ministryofjustice/community-api/pull/595

PR #593, opened 2022-02-10
PI 68: Current and Historic Order Supervisor on Community API Convictions

trozzomoj:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
```diff
@@ -36,6 +37,7 @@
 @Entity
 @Table(name = "ENFORCEMENT")
 @Where(clause = "SOFT_DELETED = 0")
+@EqualsAndHashCode(callSuper=false)
```
mxco86, line 40: Is this change related to the work?

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -81,27 +85,28 @@ public ConvictionTransformer() {
...
+                .orElse(null))
+            .inBreach(event.isInBreach())
+            .failureToComplyCount(event.getFailureToComplyCount())
+            .breachEnd(event.getBreachEnd())
+            .latestCourtAppearanceOutcome(courtAppearances.map(ConvictionTransformer::latestOutcomeOf).orElse(null))
+            .responsibleCourt(Optional.ofNullable(event.getCourt()).map(CourtTransformer::courtOf).orElse(null))
+            .courtAppearance(courtAppearances.map(CourtAppearanceBasicTransformer::latestOrSentencingCourtAppearanceOf).orElse(null))
+            .awaitingPsr(isAwaitingPsr(event.getDisposal(), courtAppearances))
+            .build();
     }
```
mxco86, line 110: I think the whole file has had it's indentation reformatted - can we revert the whitespace changes so we can see the relevant code updates?

src/main/java/uk/gov/justice/digital/delius/transformers/OrderManagerTransformer.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.Optional;
+
+public class OrderManagerTransformer {
+
+    public static OrderManager orderManagerOf(uk.gov.justice.digital.delius.jpa.standard.entity.OrderManager entity){
+        return OrderManager.builder().
+            name(Optional.ofNullable(entity.getStaff().getForename()).orElse("").
+                concat(" "+Optional.ofNullable(entity.getStaff().getSurname()).orElse(""))).
+            staffCode(Optional.ofNullable(entity.getStaff().getStaffId()).map(x->x.toString()).orElse("")).
```
mxco86, line 13: Can we name the lambda argument `staffId` rather than `x`?

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
trozzomoj: no worries.

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Enforcement.java
trozzomoj: we need to look into , definitely need to be changed

src/main/java/uk/gov/justice/digital/delius/transformers/ConvictionTransformer.java
```diff
@@ -142,6 +147,12 @@ private static KeyValue outcomeOf(StandardReference outcome) {
                 .build();
     }
 
+    private static List<OrderManager> orderManagersOf(@NotNull Event event){
+        return Optional.ofNullable(event.getOrderManagers()).map(x->x.stream().map(OrderManagerTransformer::orderManagerOf).collect(Collectors.toList())).orElse(Collections.emptyList());
```
mxco86, line 151: Can we name the lambda argument `orderManager`

src/main/java/uk/gov/justice/digital/delius/transformers/OrderManagerTransformer.java
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.justice.digital.delius.data.api.OrderManager;
+
+import java.util.Optional;
+
+public class OrderManagerTransformer {
+
+    public static OrderManager orderManagerOf(uk.gov.justice.digital.delius.jpa.standard.entity.OrderManager entity){
+        return OrderManager.builder().
+            name(Optional.ofNullable(entity.getStaff().getForename()).orElse("").
+                concat(" "+Optional.ofNullable(entity.getStaff().getSurname()).orElse(""))).
```
marcus-bcl, line 12: If the surname is empty here, then we'll end up with an extra space at the end of the name.  Maybe we should return the forenames and surname separately to keep it simple?

Also there is a `forename2` field on the `Staff` entity (like middle name), so if you stick with joining the names then that should be included too. Maybe something like this would be better?
```suggestion
            name(Stream.of(entity.getStaff().getForename(), entity.getStaff().getForename2(), entity.getStaff().getSurname())
				.filter(StringUtils::hasLength)
				.collect(joining(delimiter))).
```
(which would match the method used in the user management tool: https://github.com/ministryofjustice/ndelius-um/blob/main/src/main/java/uk/co/bconline/ndelius/util/NameUtils.java)

src/main/java/uk/gov/justice/digital/delius/transformers/OrderManagerTransformer.java
```diff
@@ -0,0 +1,20 @@
...
+            staffCode(Optional.ofNullable(entity.getStaff().getStaffId()).map(staffId->staffId.toString()).orElse("")).
+            dateStartOfAllocation(entity.getTeam().getStartDate()).
+            dateEndOfAllocation(entity.getTeam().getEndDate()).
+            officerId(entity.getOrderManagerId()).
+            probationAreaId(entity.getProbationArea().getProbationAreaId()).
+            teamId(entity.getTeam().getTeamId()).build();
```
marcus-bcl, line 18: Minor thing - any reason for putting the `.` at the end of the line here?  Other places in the project seems to put it at the start.

trozzomoj: ok not an issue , I think is just agreed style. I ll put  them on the other side.

src/main/java/uk/gov/justice/digital/delius/transformers/OrderManagerTransformer.java
trozzomoj: The requirement is just for name. I think should never be null, they are both defined as not null in the database. I have used Optional.ofNUllable just for test purpose, because in 58 tests  are passed as null. The delimiter idea is good and I guess should be an empty space . I think Name utils needs to be reviewed ;) eheh.

marcus-bcl (approved):

PR #598, opened 2022-02-23
Pi 74

trozzomoj:


src/main/java/uk/gov/justice/digital/delius/transformers/AssessmentTransformer.java
```diff
@@ -10,33 +10,34 @@
...
+                return ogrsassessmentscore;
+            }
+            if (null != ogrsassessmentscore) {
+                if (OGRS.getAssessmentDate().isAfter(oASYSAssessmentDate)) {
+                    return ogrsassessmentscore;
                 }
             }
             return null;
-        }).orElse(OASYSAssessmentScore);
+        }).orElse(oASYSAssessmentScore);
```
marcus-bcl, line 40: See what you mean now, this logic is a bit confusing - maybe worth adding a comment to say what it's doing?

My understanding is that it returns the score from either the OGRS assessment or the OASys assessment based on the following rules:
* If only an OGRS assessment is present, return that.  
* If both assessments are present, return the one with the most recent assessment date.  
* Otherwise return the score from the OASys assessment.

src/main/java/uk/gov/justice/digital/delius/transformers/AssessmentTransformer.java
```diff
@@ -10,33 +10,34 @@
...
             .builder()
             .rsrScore(offender.getDynamicRsrScore())
-            .ogrsScore(getOGRSScore(OGRSAssessment, OASYSAssessment))
+            .ogrsScore(getOGRSScore(ogrsAssessment, OASYSAssessment))
+            .orgsLastUpdate(ogrsAssessment.map(OGRSAssessment::getLastUpdatedDate).orElse(null))
             .build();
     }
 
-    private static Integer getOGRSScore(Optional<OGRSAssessment> OGRSAssessment, Optional<OASYSAssessment> OASYSAssessment) {
+    private static Integer getOGRSScore(Optional<OGRSAssessment> oGRSAssessment, Optional<OASYSAssessment> oASYSAssessment) {
```
stevomcallister, line 22: Adding a comment just to see if I can but also, shouldn't it be ogrsAssessment rather than oGRSAssessment

src/main/java/uk/gov/justice/digital/delius/transformers/AssessmentTransformer.java
trozzomoj: I did not change this logic . I have just  adjusted the  variable names to camelCase

src/main/java/uk/gov/justice/digital/delius/transformers/AssessmentTransformer.java
trozzomoj: Yep , good point , I have renamed them all so that they might look at least consistent to each other

src/main/java/uk/gov/justice/digital/delius/transformers/AssessmentTransformer.java
```diff
@@ -6,39 +6,46 @@
...
-                final var OGRSAssessmentDate = OGRS.getAssessmentDate();
-                if (OGRSAssessmentDate.isAfter(OASYSAssessmentDate)) {
-                    return OGRSAssessmentScore;
+    private static LocalDate getLastUpdateDate( Optional<OGRSAssessment> ogrsAssessment, Optional<OASYSAssessment> oasysAssessment){
+        var val1 = ogrsAssessment.map(OGRSAssessment::getLastUpdatedDate).orElse(null);
+        var val2 = oasysAssessment.map(OASYSAssessment::getLastUpdatedDate).orElse(null);
+        if (val1 != null && val2 != null) {
+            return Stream.of(val1, val2).max(LocalDate::compareTo).orElse(null);
+        }
+        return Optional.ofNullable(val1).orElse(val2);
```
marcus-bcl, line 31: Looks good, you could simplify this though - no need for null-checks as `flatMap(Optional::stream)` can do that for you e.g.

```suggestion
        var ogrsDate = ogrsAssessment.map(OGRSAssessment::getLastUpdatedDate);
        var oasysDate = oasysAssessment.map(OASYSAssessment::getLastUpdatedDate);
        return Stream.of(ogrsDate, oasysDate).flatMap(Optional::stream).max(LocalDate::compareTo).orElse(null);
```

marcus-bcl (approved):

PR #600, opened 2022-02-28
PI-89: changes to allocation dates

stevomcallister:


mxco86 (approved):
LGTM - Are we waiting for @carlov20 to confirm whether it's ok to return a DateTime?

marcus-bcl (approved):

PR #608, opened 2022-03-07
Add telephone number from LDAP to staff details endpoint

smcveigh941:
* Telephone number exists on the staff details by username endpoint, but is missing from the staff details by staffId endpoint
* This PR adds it by re-using the `addFieldsFromLdap` private function


mxco86 (approved):
LGTM

PR #609, opened 2022-03-08
PI-101 Handle null staff/team on OrderManager

marcus-bcl:
This is a rare occurrence, however there are 42 old records in production where the ALLOCATION_STAFF_ID / ALLOCATION_TEAM_ID columns are null. Hopefully the data can be corrected by NDST, but this change ensures we'll safely handle any null values in the meantime.


trozzomoj (approved):

mxco86 (approved):
👍

PR #615, opened 2022-03-17
Pi 100

trozzomoj:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/StaffRepository.java
```diff
@@ -1,21 +1,27 @@
...
-public interface StaffRepository extends JpaRepository<Staff, Long> {
+public interface StaffRepository extends JpaRepository<Staff, Long>{
     Optional<Staff> findByStaffId(Long staffId);
 
     Optional<Staff> findByOfficerCode(String officerCode);
 
-    @Query("select u.staff from User u where upper(u.distinguishedName) = upper(:username)")
+   // @Query("select st from Staff st JOIN st.teams teams where st.probationArea.code = ?1 and teams.district.borough.code = ?2 and st.grade.codeValue = ?3")
+   @Query("select st from Staff st JOIN st.teams teams where st.probationArea.code = ?1 and teams.localDeliveryUnit.code = ?2 and st.grade.codeValue = ?3")
+   List<Staff> findStaffByProbationAreaAndPduCodeAndGrade(String probationAreaCode, String lduCode, String gradeCode);
```
marcus-bcl, line 22: Don't think you need this repository method any more?

src/testIntegration/java/uk/gov/justice/digital/delius/jpa/standard/repository/StaffRepositoryTest.java
```diff
@@ -30,4 +32,17 @@ public void canSeeTheStaff() {
...
+    public void findHeadsOfPDU(){
+       var result =  staffRepository.findStaffByProbationAreaAndPduCodeAndGrade("N07","N07NPS1", "NPSB");
+        assertThat(result.size() == 2);
+    }
+
+    @Test
+    public void ShouldReturnEmptyCollectionWhenNotFound(){
+        var result =  staffRepository.findStaffByProbationAreaAndPduCodeAndGrade("N0711","N07NPS1", "NPSB");
+        assertThat(result.size() == 0);
+    }
```
marcus-bcl, line 46: Same here, I think you can remove these tests.

marcus-bcl (commented):
Minor comment - looks good otherwise though 👍

src/main/java/uk/gov/justice/digital/delius/controller/secure/StaffResource.java
```diff
@@ -182,4 +182,10 @@ public Caseload getCaseloadForStaff(
             .orElseThrow(() -> new NotFoundException(String.format("Staff member with code %s", staffCode))));
     }
 
+    @ApiOperation(value = "Return the list of heads of a specific probation delivery unit (aka borough)")
+    @GetMapping(path = "/staff/pduHeads/{pduCode}")
```
stevomcallister, line 186: picky one... should this have a capital H? I think normally we use dashes so pdu-heads

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Borough.java
```diff
@@ -71,4 +83,9 @@
     @JoinColumn(name = "BOROUGH_ID")
     private List<District> districts;
 
+    @ManyToMany
+    @JoinTable(name = "R_LEVEL_2_HEAD_OF_LEVEL_2",
+        joinColumns = @JoinColumn(name = "BOROUGH_ID"),
```
timharrison-moj, line 88: Is R_LEVEL_2_HEAD_OF_LEVEL_2 a database view containing staff at the grade that are PDU heads? i.e. PSA, PSB, PSC.

timharrison-moj (commented):
LGTM. Thanks.
I'm interested in how the table/view `r_level_2_head_of_level_2` is populated and maintained as staff are promoted / join or leave as PDU heads?

src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/Borough.java
marcus-bcl: The table was added recently in Delius to allow authorised users to explicitly mark who are the heads of the PDU, regardless of staff grade.  Among other reasons, I believe this was so staff could be temporarily added while people were on leave etc - without having to change people's grades. We originally went down the route of looking up by PSB grade, until we discovered this Delius change which seems more appropriate.

It's worth noting however that this change is not currently live.  It's in all the lower environments and is planned to go-live next weekend (26th March).  As part of the go-live, they'll be running a script to populate the table based on information directly from the PDUs, so it should be accurate.  They've already tested this in Pre-Prod, so feel free to check that the data in there matches what you're after.

timharrison-moj: Thanks Marcus. All good then.

src/main/java/uk/gov/justice/digital/delius/controller/secure/StaffResource.java
trozzomoj: I do agree on this, but I think they are all like this in community api. Otherwise yes defo.

marcus-bcl:
We might need to hold off on merging this (or somehow feature flagging the entity relationship?), until 26th March when the table is added to the Delius database.

mxco86:
> LGTM. Thanks. I'm interested in how the table/view `r_level_2_head_of_level_2` is populated and maintained as staff are promoted / join or leave as PDU heads?

This will be managed by a new screen in NDelius. It should be kept up-to-date and correct as it will be used for operational contacts

marcus-bcl (approved):

PR #646, opened 2022-04-14
Define community-api for developer portal

sldblog:
Adds a Backstage entity metadata file to this repository so that the
component can be added to the developer portal
(https://developer-portal.apps.live.cloud-platform.service.justice.gov.uk/catalog)

Strategically, this file will replace most of https://github.com/ministryofjustice/hmpps-architecture-as-code/blob/main/src/main/kotlin/uk/gov/justice/hmpps/architecture/model/Delius.kt

Similar:

- https://github.com/ministryofjustice/hmpps-delius-api/pull/275

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: hmpps-community-api
+  title: Community API service
+  description: API service over the NDelius DB used by HMPPS Digital team applications and services
+  tags:
+    - java
+    - spring-boot
+spec:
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,36 @@
...
+  name: hmpps-community-api
+  title: Community API service
+  description: API service over the NDelius DB used by HMPPS Digital team applications and services
+  tags:
+    - java
+    - spring-boot
+spec:
+  type: service
+  lifecycle: production
+  owner: group:delius-api
```
sldblog, line 14: Is this the correct github team as owner?

PR #657, opened 2022-05-18
PI-178

stevomcallister:


src/main/java/uk/gov/justice/digital/delius/utils/ContentDispositionHeader.java
```diff
@@ -0,0 +1,23 @@
...
+
+import lombok.Getter;
+
+import java.nio.charset.StandardCharsets;
+
+import static java.net.URLEncoder.encode;
+
+
+@Getter
+public class ContentDispositionHeader {
```
marcus-bcl, line 11: Did you look into using the [org.springframework.http.ContentDisposition](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/ContentDisposition.Builder.html) class for this?  It provides a builder pattern and seems to do most of what you're doing here out of the box.

marcus-bcl (commented):
Should we add a test that verifies it can handle filenames with special characters?

src/testIntegration/java/uk/gov/justice/digital/delius/service/AlfrescoServiceTest.java
```diff
@@ -37,14 +37,18 @@ public void shouldBeNotFoundIfDocumentDoesNotBelongToOffender() {
...
+        AlfrescoExtension.alfrescoMockServer.stubDetailsSuccess("123", "T1234", filename);
         AlfrescoExtension.alfrescoMockServer.stubFetchDocument("123", new byte[]{'a', 'b', 'c'});
 
         final var response = alfrescoService.getDocument("123", "T1234");
 
         assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
         assertThat(response.getBody().contentLength()).isEqualTo(3);
-        assertThat(response.getHeaders().getFirst(HttpHeaders.CONTENT_DISPOSITION)).isEqualTo("attachment; filename=\"document.pdf\"");
+        assertThat(response.getHeaders().getFirst(HttpHeaders.CONTENT_DISPOSITION)).isEqualTo(ContentDisposition.attachment()
+            .filename(filename, StandardCharsets.UTF_8).build().toString());
```
marcus-bcl, line 51: Tiny nitpick - would it make more sense to assert against the actual expected string?  Just to make it clear that the values will be encoded.
```suggestion
        assertThat(response.getHeaders().getFirst(HttpHeaders.CONTENT_DISPOSITION))
            .isEqualTo("attachment; filename*=UTF-8''document%20%281%29.pdf");
```

marcus-bcl (approved):

mxco86 (dismissed):

marcus-bcl (approved):

PR #688, opened 2022-07-05
Pic 2262

rschallapalli:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/CourtReport.java
```diff
@@ -131,4 +131,8 @@ public class CourtReport {
 
     @Column(name = "PENDING_TRANSFER")
     private Long pendingTransfer;
+
+    @JoinColumn(name = "DELIVERED_REPORT_REASON_ID", referencedColumnName = "STANDARD_REFERENCE_LIST_ID", insertable = false, updatable = false)
```
jhevans, line 135: I think this is the wrong column, it should be `DELIVERED_COURT_REPORT_TYPE_ID` referencing `R_COURT_REPORT_TYPE` shouldn't it?

jhevans (changes requested):
@rschallapalli  I think there's a mistake in the column references which needs checking, otherwise good.

jhevans (approved):

jhevans (approved):

PR #706, opened 2022-08-17
MAS-457 added check for latest tier before save

carlov20:


src/main/java/uk/gov/justice/digital/delius/service/TierService.java
```diff
@@ -42,13 +42,16 @@ public void updateTier(String crn, String tier) {
         final var offender = getOffender(crn, telemetryProperties);
         final var offenderId = offender.getOffenderId();
         final var changeReason = getChangeReason(telemetryProperties);
+        final var currentTier = managementTierRepository.findFirstByIdOffenderIdOrderByIdDateChangedDesc(offenderId);
         final var updatedTier = getUpdatedTier(tier, tierWithUPrefix, telemetryProperties);
 
-        writeTierUpdate(updatedTier, offenderId, changeReason);
-        writeContact(offender, changeReason, updatedTier, telemetryProperties);
-        spgNotificationService.notifyUpdateOfOffender(offender);
+        if(updatedTier != currentTier.getId().getTier()) {
```
AndySweilam, line 48: Null pointer exception possible?

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_updateTier.java
```diff
@@ -56,6 +57,28 @@ public void createsContactWhenTierUpdated() {
         List<Contact> updatedContacts = contactService.contactsFor(2500343964L, contactFilter);
 
         assertThat(updatedContacts.stream().anyMatch(c ->  c.getNotes().contains("Tier: B-1"))).isTrue();
+        managementTierRepository.deleteById(tierB1.getId());
+    }
+
+
```
AndySweilam, line 63: another test to bring in a tier when it is the first tier calculation.

AndySweilam (changes requested):

src/testIntegration/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource_updateTier.java
carlov20: added a test for it, you were correct, null was possible

src/main/resources/db/data/V1_37__add_staff_for_no_tier.sql
```diff
@@ -0,0 +1,11 @@
+Insert into TEAM (TEAM_ID, CODE, DESCRIPTION, DISTRICT_ID, LOCAL_DELIVERY_UNIT_ID, TELEPHONE, UNPAID_WORK_TEAM, ROW_VERSION,
```
markreesmoj, line 1: Is this insert necessary? The tests don't seem to refer to this

carlov20: They are needed to create the contact https://github.com/ministryofjustice/community-api/blob/main/src/main/java/uk/gov/justice/digital/delius/service/TierService.java#L56...L57

src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/ManagementTierRepository.java
```diff
@@ -7,4 +7,6 @@
 
 @Repository
 public interface ManagementTierRepository extends JpaRepository<ManagementTier, ManagementTierId> {
+
+    ManagementTier findFirstByIdOffenderIdOrderByIdDateChangedDesc(Long offenderId);
```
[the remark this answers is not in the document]
carlov20: Tried this and get:
```
Caused by: org.springframework.data.mapping.PropertyReferenceException: No property 'offenderId' found for type 'ManagementTier'
```
I believe the Id is needed because its an embedded ID and the offenderId and dateChanged fields are within the id field in the entity https://github.com/ministryofjustice/community-api/blob/main/src/main/java/uk/gov/justice/digital/delius/jpa/standard/entity/ManagementTier.java#L33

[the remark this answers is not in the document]
markreesmoj: Andy and I discussed this yesterday and reached the same conclusion

src/main/java/uk/gov/justice/digital/delius/service/TierService.java
```diff
@@ -33,7 +33,6 @@ public class TierService {
     private final ContactService contactService;
     private final StaffRepository staffRepository;
     private final TeamRepository teamRepository;
-    private final SpgNotificationService spgNotificationService;
```
markreesmoj, line 36: Why has this service and the call to it been removed?

carlov20: https://github.com/ministryofjustice/community-api/pull/706#discussion_r948167991

src/main/resources/db/data/V1_37__add_staff_for_no_tier.sql
markreesmoj: Is it because you are using this offender in a tiering test for the first time? X330899
If so makes sense

src/main/java/uk/gov/justice/digital/delius/service/TierService.java
markreesmoj: 😄

src/main/resources/db/data/V1_37__add_staff_for_no_tier.sql
carlov20: This happened when adding the no tier exists test, X330899 has got a tier already so used X320741 as it doesn't have an existing tier

PR #732, opened 2022-10-10
Switch to multi platform builds for Docker

pezholio:
This will switch to multi-platform builds for Docker, so the community-api image works on Apple Silicon Macs as well. Currently, we have issues with migrations taking a long time to run or not at all when spinning up on newer Macs.

I’ve used https://github.com/ministryofjustice/hmpps-auth/pull/1275 as inspiration, but I noticed that Community API uses `jib` to build/publish Docker images, so can’t use the shared `hmpps/build_multiplatform_docker` job from the HMPPS Circle CI orb.

Hopefully I've got all this right - please let me know if not!


pezholio:
@anthony-britton-moj Ah, that's a good point. It will also need the upstream https://github.com/ministryofjustice/hmpps-probation-integration-services/tree/main/tools/eclipse-temurin/17-jre Docker image to be tweaked to be multiplatform too

pezholio:
Have now updated the PR to use the new `eclipse-temurin:19-jre-jammy` mutliplatform build from https://github.com/ministryofjustice/hmpps-probation-integration-services

PR #743, opened 2022-10-26
Pass down referral URN as NSI external reference

sldblog:
## What does this pull request do?

Pass down referral URN as NSI external reference.

- https://github.com/ministryofjustice/hmpps-delius-api/pull/424 has exposed a new `externalReference` field.
- This pull request populates that field when new referrals are made for "Refer and monitor an intervention".

## What is the intent behind these changes?

Assign a unique "owner" identifier to each NSI created by and for interventions.

This can later be used to:

- remove the "fuzzy" matching logic for NSIs
- remove the "notes containing" matching logic for NSIs
- and replace the above with a direct lookup, e.g. `WHERE nsi.external_reference = "urn:hmpps:interventions-referral:c7aa7e1f-3fec-4ffa-b99d-3a898f7e7b4b"`

Which will be a significant simplification.

There is also the option to treat NSIs with "external references" as managed by 3rd party systems and prevent them from being mutated outside those 3rd party systems.


sldblog:
[Reduce NSI not found errors](https://trello.com/c/PoWz8WQh/546-reduce-nsi-not-found-errors)

sldblog:
Additionally, we will gain the option to prevent multiple NSIs created concurrently for the same URN.

src/test/java/uk/gov/justice/digital/delius/transformers/ReferralTransformerTest.java
```diff
@@ -5,33 +5,33 @@
...
         assertThat(prefixReferralStartNotesWithUrn("some notes", referralId))
-            .isEqualTo("urn:hmpps:interventions-referral:" + referralId.toString() + "\nsome notes");
+            .isEqualTo("urn:hmpps:interventions-referral:" + referralId + "\nsome notes");
     }
 
     @Test
-    public void doesNotPrefixesNotesWithUrn_WhenReferralIdNotSupport() {
-
-        final var referralId = UUID.randomUUID();
+    public void doesNotPrefixesNotesWithUrn_WhenReferralIdIsMissing() {
```
sldblog, line 33: (not a blocker)
As far as I know, we always have a referral UUID. @mxco86 Do you remember why we have this case?

sldblog:
Thank you @marcus-bcl! I can't merge the PR, could you merge it on my behalf, please? 🙇‍♂️

PR #747, opened 2022-10-27
PI-489 Update security scan schedule and remove Gradle OWASP check

marcus-bcl:


marcus-bcl:
> Why has the logging been removed?

It flags up a log injection error in Veracode, because an attacker can pass in parameters to the endpoints with CRLF / new line characters and create some dodgy log entries.   Also, we get all the call information from App Insights now - hopefully people aren't still looking at the logs to find out which endpoints are being called.

PR #772, opened 2022-11-25
PI-586 - fix for slow query

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/delius/service/CaseloadService.java
```diff
@@ -38,11 +39,11 @@ public Optional<Caseload> getCaseloadByStaffIdentifier(final long staffIdentifie
     public Optional<Caseload> getCaseloadByTeamCode(
         final String teamCode, final Pageable pageable, final CaseloadRole... roles
     ) {
-        if (!teamRepository.existsByCode(teamCode)) return Optional.empty();
```
marcus-bcl, line 41: You might also be able to remove the existsByCode method from the interface now.

marcus-bcl (approved):

PR #798, opened 2023-01-20
PI-728 add consideration of event and disposal being active

anthony-britton-moj:


src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/CaseAllocationRepository.java
```diff
@@ -9,5 +10,14 @@
 
 @Repository
 public interface CaseAllocationRepository extends JpaRepository<CaseAllocation, Long> {
-    Optional<CaseAllocation> findFirstByOffenderIdAndAllocationDecisionDateNotNullOrderByAllocationDecisionDateDesc(@Param("offenderId") Long offenderId);
+    @Query("select ca from CaseAllocation ca " +
```
marcus-bcl, line 13: Is it worth still selecting the first, just in case there are duplicates that are currently being handled by the `findFirstBy`?

stevomcallister (approved):

marcus-bcl (approved):

PR #822, opened 2023-03-28
Add soft deleted check to find offender by crn

anthony-britton-moj:
Prepared in case it is just soft deleted check missing - waiting for confirmation before merge


marcus-bcl:
Is there a chance this might break things for other clients?  Carlo's issue isn't related to multiple offenders with the same CRN, so might be safer to leave this as is till we know what it is.

PR #823, opened 2023-03-28
PI-949 Simplify SQL for userAccess endpoint

marcus-bcl:


marcus-bcl:
> > > Shouldn't we be moving to the new access endpoint in the workforce api at this point?
> > 
> > 
> > Yes workforce should be moving over anyway, but this is the most heavily used Community API endpoint - so it makes sense to optimise it for the other clients, and ease the impact on the DB.
> 
> Whilst we `ease` the load on the db though - we are removing the `carrot` for people adopting the other services and moving off Community API - bug fix is one thing - this is rewriting the endpoint

Sure, good point - I'll hold off on merging for now and can have a chat about it on stand up tomorrow.

